PR #16, opened 2017-09-19
	fix duplicate group id issue

ericbroder:


pages/models.py
```diff
@@ -609,6 +609,20 @@ def add_group_view(self, request):
 
         return render(request, 'pages/add_group.html', {'form': form})
         
+    '''
+    Get new group id that is random and not in use
+    '''
+    def get_new_group_id(self):
+        '''
+        Find random integer between 1000 and 9998 and not in use by another group        
+        TODO: more scalable solution that doesn't use random guess approach and
```
ericbroder: @cjmabry Here's what a more scalable solution could be:

- Find list of all group ids in use
- Create list of numbers from 1000 to 9998 and filter out group ids in use
- Take remaining unused numbers and randomly pick one
- Display user friendly error message if there are no unused numbers left to choose from

Do you think we should implement a better solution now, or make a ticket to do it later?

ericbroder: Basically this changeset fixes immediate problem but is not a great long term solution.

cjmabry: Let's go ahead an merge this because we're seeing some errors related to this on occasion, and I'll add create more scaleable solution to the backlog.

PR #17, opened 2017-09-28
add notes field to groups table

ericbroder:
@cjmabry is there anything else I need to do? seems to be working ok locally.


local_groups/models.py
```diff
@@ -10,37 +10,37 @@
 from django.contrib.gis.db.models import PointField
 
 
-class Group(models.Model):
```
ericbroder: not sure why getting so many whitespace changes. something to do with atom maybe.

cjmabry: No whitespace on empty lines is preferred, so your editor is correct here. Someone might have been using improperly configured editor before.

ericbroder: Ok cool 👍

PR #19, opened 2017-09-28
add election dates to candidate application admin page

ericbroder:


nominations/admin.py
```diff
@@ -108,7 +108,23 @@ class ApplicationAdmin(admin.ModelAdmin):
     # needed for ordering as readonly_fields are automatically listed at the end
     fields = ('submitted_dt','group','rep_email','rep_first_name','rep_last_name','rep_phone','candidate_first_name','candidate_last_name','nomination','questionnaire','candidate_office','candidate_district','candidate_city','candidate_state','authorized_email','vol_incumbent','vol_dem_challenger','vol_other_progressives','vol_polling','vol_endorsements','vol_advantage','vol_turnout','vol_win_number','vol_fundraising','vol_opponent_fundraising','vol_crimes','vol_notes','status',)
 
-    readonly_fields = ('submitted_dt','rep_email','group','rep_first_name','rep_last_name','rep_phone','candidate_first_name','candidate_last_name','candidate_office','candidate_district','candidate_city','candidate_state','authorized_email')
+    readonly_fields = (
```
cjmabry: Thanks for breaking up long lines. If you're using Atom IDE with python package you'll get style errors, but if not, we should be moving to https://www.python.org/dev/peps/pep-0008/ as we make commits. Woo!

PR #20, opened 2017-10-02
As a staff member, I want to generate a one-page summary PDF of an endorsement to share with the board and speed up endorsements. 

cjmabry:


cjmabry:
[(8) As a staff member, I want to generate a one-page summary PDF of an endorsement to share with the board and speed up endorsements. [8]](https://trello.com/c/065rc5NV/2-8-as-a-staff-member-i-want-to-generate-a-one-page-summary-pdf-of-an-endorsement-to-share-with-the-board-and-speed-up-endorsement)

nominations/admin.py
```diff
@@ -166,8 +295,6 @@ def get_form(self, request, obj=None, **kwargs):
             'candidate_district',
             'candidate_city',
             'candidate_state',
-            'get_general_election',
-            'get_primary_election',
```
ericbroder: Is this accidental? Looks like a messed up merge?

nominations/views.py
```diff
@@ -486,12 +488,30 @@ def form_valid(self, form):
...
         context_data = super(CreateInitiativeView, self).get_context_data(*args, **kwargs)
         context_data['user'] = self.request.session['profile']
         return context_data
+
+
+
+class ApplicationPDFView(PDFTemplateView):
+
+    template_name = 'admin/application_pdf.html'
+    download_filename = 'hello.pdf'
```
ericbroder: Does this do anything?

ourrevolution/settings.py
```diff
@@ -77,16 +77,17 @@
...
     # Group Goodies
     'phonenumber_field',
     'multiselectfield',
     'django_countries',
     'recurrence',
     'address',
-    
     'bootstrap3',
     'crispy_forms',
+    'ckeditor',
```
ericbroder: Why do we need a new editor?

requirements.txt
```diff
@@ -66,3 +67,4 @@ wcwidth==0.1.7
 webencodings==0.5
 whitenoise==3.2.2
 wsgiref==0.1.2
+xhtml2pdf==0.2b1
```
ericbroder: Are we sure production servers are set up to support these pdf libraries?

nominations/admin.py
cjmabry: It is accidental but it doesn't affect functionality because we're using fieldsets here now. Will just remove this part, thanks for flagging.

ourrevolution/settings.py
cjmabry: This is for rich text, WYSIWYG for staff to add more than just text. Mostly this is for list functionality because this will be used to generate the candidate write-up.

requirements.txt
cjmabry: Yup, should be fine.

nominations/views.py
cjmabry: Nope, removed.

nominations/templates/admin/application_pdf.html
```diff
@@ -0,0 +1,182 @@
+{% extends "easy_pdf/base.html" %}
+
+
+
+{% block content %}
+    {% load static %}
+    <link rel="stylesheet" href="{% static "dist/css/main.min.css" %}?v=122">
```
ericbroder: Do we have to manually version the css files with url params? This is not automated?

nominations/templates/admin/application_pdf.html
```diff
@@ -0,0 +1,182 @@
...
+
+    <style>
```
ericbroder: Suggest to put this stuff in css files unless there is a need to have it here.

nominations/templates/admin/application_pdf.html
```diff
@@ -0,0 +1,182 @@
...
+            {% if app.vol_opponent_fundraising %}
+              <strong>Opponent Fundraising:</strong> ${{app.vol_opponent_fundraising}}
+            {% endif %}
+          </li>
+        </ul>
+      {% endif %}
+
+      {% if app.local_group_info %}
+        <h2>{{app.group}} Info</h2>
+        {{app.local_group_info|safe}}
```
ericbroder: Why using `|safe` everywhere? What does it do?

nominations/templates/admin/nominations/change_form.html
```diff
@@ -0,0 +1,94 @@
...
+    {% include inline_admin_formset.opts.template %}
+{% endfor %}
+{% endblock %}
+
+{% block after_related_objects %}{% endblock %}
+
+{% block submit_buttons_bottom %}{% submit_row %}{% endblock %}
+
+{% block admin_change_form_document_ready %}
+    <script type="text/javascript">
```
ericbroder: Suggest move this to js file if possible.

nominations/migrations/0056_auto_20171002_0040.py
```diff
@@ -0,0 +1,55 @@
...
+        ),
+        migrations.AddField(
+            model_name='application',
+            name='local_group_info',
+            field=models.TextField(blank=True, null=True, verbose_name='OR Local Group Info:'),
+        ),
+        migrations.AddField(
+            model_name='application',
+            name='local_support',
+            field=models.TextField(blank=True, null=True, verbose_name='Notes or Flags:'),
```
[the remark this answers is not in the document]
ericbroder: Oh I see it's fixed in later migration.

nominations/templates/admin/application_pdf.html
cjmabry: Correct. Ticket to fix here https://trello.com/c/D76LqScs/9-implement-automated-hashing-for-static-assets

nominations/templates/admin/application_pdf.html
cjmabry: There was some CSS weirdness with this PDF generator where in-code styles were working and linked files were not.

nominations/templates/admin/nominations/change_form.html
cjmabry: This is the default django template, copied and overridden as seen here https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#overriding-admin-templates, so it doesn't make sense to move to external JS in this context.

nominations/templates/admin/application_pdf.html
cjmabry: These are rich text fields, consisting of HTML, i.e. `<p>Some text</p>`. The `safe` filter is a django template filter that disables HTML escaping. https://docs.djangoproject.com/en/1.8/ref/templates/builtins/#safe. Without it, we'd get `<p>Some text</p>` in the PDF instead of `Some text`.

PR #21, opened 2017-10-02
rename field for other candidates

ericbroder:


nominations/migrations/0056_auto_20171002_2054.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
ericbroder: Same migration prefix 0056_* as another PR https://github.com/Our-Revolution/site/pull/20/files#diff-6e73b9542001040d0afd65bc23e2bf14R1

Is this something we need to address or avoid? Or do we not care if prefix is the same as long as the full timestamp is different?

ericbroder:
@cjmabry with added help text now

nominations/models.py
```diff
@@ -178,7 +178,13 @@ class Application(models.Model):
     # Volunteer Data Entry
     vol_incumbent = models.NullBooleanField(null=True, blank=True, verbose_name='Incumbent?')
     vol_dem_challenger = models.NullBooleanField(null=True, blank=True, verbose_name='If primary, who are the Democratic challengers?')
-    vol_other_progressives = models.TextField(null=True, blank=True, max_length=500, verbose_name='Other progressives running:')
+    vol_other_progressives = models.TextField(
```
cjmabry: We should actually rename this field to vol_other_candidates so it is descriptive of the data inside. We'll then need to change the two instances where vol_other_progressives is used in admin.py. A new migration will handle this rename.

nominations/models.py
```diff
@@ -178,7 +178,13 @@ class Application(models.Model):
...
+        null=True,
+        blank=True,
+        max_length=500,
```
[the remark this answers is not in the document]
cjmabry: This has been resolved.

nominations/models.py
ericbroder: @cjmabry my concern is that im not sure right way to do this with our infrastructure. given we have a/b deploy, does this mean we need to keep the old field name available for now and then completely remove it in future deploy? sometimes field renames are tricky if one server deploys new code and migrations but there is another server still running app on old code.

also another issue is that i think rename will drop that data from column, so may need to create new column first, copy data from old column to new column, and then drop old column? maybe you could point me to code example if we have done this before.

nominations/models.py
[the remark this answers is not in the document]
ericbroder: Erika confirmed 500.

ericbroder:
@cjmabry we discussed this. added code comments and otherwise will merge as is.

PR #22, opened 2017-10-03
move submitted at field to left and add admin css with compact tables

ericbroder:


ourrevolution/settings.py
```diff
@@ -38,9 +38,20 @@
 
 
 # Application definition
-
+'''
+When several applications provide different versions of the same resource
```
ericbroder: Figuring out this step was really confusing so I hope this is correct. I could not get admin template override to work until I did this part.

pages/static/dist/css/admin.css
```diff
@@ -0,0 +1,8 @@
+#result_list th, #result_list td {
```
ericbroder: Why are we storing the compiled version in git? Shouldn't this be built on the fly and not stored in version control? Don't we get merge conflicts by storing the compiled versions in git?

pages/templates/admin/base_site.html
```diff
@@ -0,0 +1,15 @@
+{% extends "admin/base.html" %}
```
ericbroder: Ditto I hope this is right, took me a while to get this working.

gulpfile.js
```diff
@@ -58,7 +58,10 @@ function bundle() {
 
 /* Compile SASS */
 gulp.task('sass', function() {
-  return gulp.src('pages/static/src/scss/main.scss')
+  return gulp.src([
+    'pages/static/src/scss/main.scss',
+    'pages/static/src/scss/admin.scss'
```
ericbroder: Does this look right? Want to process both files in similar way but don't want to merge them together.

cjmabry: Yup, this works!

pages/static/dist/css/admin.css
cjmabry: It's a holdover from our old infrastructure setup that we eventually need to fix. But yes, you're correct, we should not be storing compiled assets in git.

pages/templates/admin/base_site.html
cjmabry: I'm getting these styles applied to every admin model. Very close, but it's a weird thing about django admin overrides.

We need to override just this template for this particular model and no others. The template that shows the list of models in the admin is called `change_list.html`, and it's one of five that can be overridden on a per model basis (https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#templates-which-may-be-overridden-per-app-or-model).

To override this template, we need to place our new template file in the proper directory for django to recognize it as this model's override (https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#overriding-admin-templates).

So, we need to copy the change_list.html template from Django 1.8 (https://github.com/django/django/blob/stable/1.8.x/django/contrib/admin/templates/admin/change_list.html), override it as we need (throw in the admin style you created), and Django will automatically apply it to the proper model if we have it in the right directory.

This new template should reside in `nominations/templates/admin/nominations/change_list.html`.

Let me know if you have questions.

ourrevolution/settings.py
cjmabry: If we override the template on a per model basis (https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#templates-which-may-be-overridden-per-app-or-model), we shouldn't need to change app order.

ericbroder: My intention was to override for the whole admin theme, not a single model.

pages/templates/admin/base_site.html
ericbroder: I considered altering the theme just for a specific model, but I thought applying the changes for the whole admin portal would be better UX since it's more consistent. Do you really think it's better to have a more compact theme just for a single model admin page, and have the default version elsewhere? I figured more compact everywhere in general would be better but either way is fine with me.

ourrevolution/settings.py
cjmabry: Hmmmm. I'm fine with this. If we get complaints we can override per model. Disregard, but hey, at least you know all about overriding per model now haha.

pages/templates/admin/base_site.html
cjmabry: I'm cool with this now that I think about it. We can change if we get negative feedback.

ourrevolution/settings.py
ericbroder: haha actually I figured out how to do it per model first, and then I spent extra time figuring out how to do it for the whole admin theme. So either way was a good learning experience 😄

PR #36, opened 2017-10-31
TECH-255 fix search by name

ericbroder:


pages/templates/pages/election_tracking_page.html
```diff
@@ -311,12 +311,13 @@
...
             })
-            
+
             var sortingBy = "state";
             var sortingAsc = true;
-            
+
             var options = {
                 valueNames: [
+                  'name',
```
ericbroder: I think this will fix but I can't test locally. Was able to do limited testing on prod by editing code in my browser.

ericbroder: yes

PR #37, opened 2017-11-01
TECH-355 add notification banner to cms

ericbroder:


pages/models.py
```diff
@@ -58,6 +59,27 @@ class BasePage(Page):
             ImageChooserPanel('social_image')
         ]
 
+@register_snippet
+@python_2_unicode_compatible  # provide equivalent __unicode__ and __str__ methods on Python 2
```
ericbroder: Not really sure what this does - copied from http://docs.wagtail.io/en/v1.10.1/topics/snippets.html#snippet-models

pages/models.py
```diff
@@ -58,6 +59,27 @@ class BasePage(Page):
...
+class NotificationBanner(models.Model):
```
ericbroder: Does this look like right model/fields that we want?

pages/templates/pages/tags/notification-banner.html
```diff
@@ -0,0 +1,15 @@
+<!-- notification banner -->
+{% for notification in notifications %}
+    <div class="notification-banner">
```
ericbroder: Is this template sufficient or do we need more dynamic fields in here?

pages/templatetags/pages_tags.py
```diff
@@ -0,0 +1,12 @@
+from django import template
+from pages.models import *
+
+register = template.Library()
+
+# Notification Banner snippets that are set to show
+@register.inclusion_tag('pages/tags/notification-banner.html', takes_context=True)
+def notification_banner(context):
+    return {
+        'notifications': NotificationBanner.objects.filter(show=True),
```
cjmabry: Could we make this get the latest notification banner? Perhaps with https://docs.djangoproject.com/en/1.8/ref/models/querysets/#latest

I want to avoid someone adding two banners on accident if we could, or only allow creation of one notification banner.

ericbroder: Ok sure, will just take first record from list then. So as long as 0 or 1 record is set to show then will work as expected. And if more than 1 is set to show will still only display 1.

cjmabry:
Sorry should have specified, can we return the last item and not the first? We can just use last() where you used first() (https://docs.djangoproject.com/en/1.11/ref/models/querysets/#last). As is, if new notification banners are added, we still show the first one. Trying to think about how comms team will use, I think this might be confusing for them, as they will add a notification banner and old one will still show. Thoughts?

ericbroder:
@cjmabry I think the difference b/w `first()` or `last()` is trivial in this case. There should only be 0 or 1 notification banner set to show. If more than 1 banner is set to show, then it is being used incorrectly, and it will be obvious how to fix, by disabling the show field on any banner which is displayed that we do not want to display. If "they will add a notification banner and old one will still show", then they just need to mark the old banner to not show. I think this makes intuitive sense from user perspective.

On the other hand if we support the idea that they can just add a new banner and it automatically replaces the old banner, that is not good because when they disable new banner then the old banner will show up again. It is better to disable the old banner explicitly.

On the other hand, we could potentially automate solution so that if a banner is set to show, then automatically all others are disabled from showing. But that seems like probably not worth the development effort.

PR #38, opened 2017-11-02
TECH-432 fix meta image for candidate pages

ericbroder:


pages/templates/pages/candidate_endorsement_page.html
```diff
@@ -8,7 +8,7 @@
     <meta property="og:description" content="The next step for Bernie Sanders' movement is a new group called Our Revolution, which will fight to transform America and advance the progressive agenda that we believe in." />
     <meta property="og:url" content="{{ request.url }}" />
     <meta property="og:site_name" content="Our Revolution" />
-    <meta property="og:image" content="https://ourrevolution.com/img/fb.png" />
+    <meta property="og:image" content="{{ page.candidate.photo.url }}" />
```
ericbroder: Do you know if this will output relative or absolute urls? I'm seeing inconsistent results. In my local env all these images are broken because the relative url doesnt work. In prod apparently it goes to correct full url in s3.

ericbroder: Ok I found the logic in code, nevermind.

PR #40, opened 2017-11-05
TECH-428 add 2017 initiative results

ericbroder:


pages/templates/pages/election_tracking_page.html
```diff
@@ -186,17 +186,20 @@
...
-                                                            {% elif initiative_race.initiative_votes > initiative_race.opponent_votes %}
-                                                                {{ initiative.name }} is leading {{ initiative_race.initiative_votes_percentage | urlencode }} to {{ initiative_race.opponent_votes_percentage | urlencode}}. Track this race and other progressive's results nationwide LIVE.
-                                                            {% else %}
-                                                                {{ initiative.name }} is trailing {{ initiative_race.opponent_votes_percentage | urlencode }} to {{ initiative_race.initiative_votes_percentage | urlencode }}. Track this race and other progressive's results nationwide LIVE.
-                                                            {% endif %}{% endspaceless %}&url=https://ourrevolution.com/election-2016" class="social-icon twitter-icon twitter-bg">Tweet</a>
-                                         </div>
+                                          {% comment %}
```
cjmabry: No need to keep this in commented, we can remove.

ericbroder: I suggest we leave it in for now and eventually re-enable it with updated content. I think it's a nice feature to be able to share results, these are just out of date. Instead of 2016 livetracker, could just be permanent share result feature?

cjmabry: Sure.

PR #41, opened 2017-11-07
TECH-405 add page link to election result

ericbroder:


pages/models.py
```diff
@@ -322,6 +322,12 @@ class CandidateRace(models.Model):
             ('lose', 'Lose'),
         )
     candidate = models.ForeignKey('endorsements.Candidate', null=True, blank=True, on_delete=models.SET_NULL)
+    candidate_endorsement_page = models.ForeignKey(
```
cjmabry: So problem here is that we'll have to manually set pages for over 100 races, as they do not auto-populate. What I recommend instead is accessing the candidate slug on the template side like `<a href="/candidates/{{ candidate_race.candidate.slug }}">` and avoiding any model changes or migrations.

pages/models.py
```diff
@@ -369,6 +379,12 @@ class InitiativeRace(models.Model):
             ('lose', 'Lose'),
         )
     initiative = models.ForeignKey('endorsements.Initiative', null=True, blank=True, on_delete=models.SET_NULL)
+    initiative_endorsement_page = models.ForeignKey(
```
cjmabry: Same as above but with initiatives.

pages/templates/pages/election_tracking_page.html
```diff
@@ -95,28 +95,45 @@
...
-                                                                  {{ candidate.name }} is leading {{ candidate_race.candidate_votes_percentage | urlencode }} to {{ candidate_race.opponent_votes_percentage | urlencode }}. Track this race and other progressive's results nationwide LIVE.
-                                                              {% else %}
-                                                                  {{ candidate.name }} is trailing {{ candidate_race.opponent_votes_percentage | urlencode }} to {{ candidate_race.candidate_votes_percentage | urlencode }}. Track this race and other progressive's results nationwide LIVE.
-                                                              {% endif %}{% endspaceless %}&url=https://ourrevolution.com/election-2016" class="social-icon twitter-icon twitter-bg">Tweet</a>
-                                                      </div>
-                                                    {% endcomment %}
+                                                {% if candidate_race.candidate_endorsement_page %}
```
cjmabry: There's no visual indication to the user that these are links. I recommend we link the entire card and add hover effect seen on candidate page, or perhaps make text blue and add a `glyphicon glyphicon-chevron-right` as a visual indicator.

pages/models.py
ericbroder: Unfortunately slug is not reliable for all cases. I suggest handling the backfill of old results in a future issue. The most important races are the new ones, and we can set them as we add them going forward. I agree slug is easy and automatic but some of them will be wrong urls because candidate slug is separate field in db from page url.

pages/models.py
ericbroder: Ditto about not having reliable slug.

pages/templates/pages/election_tracking_page.html
ericbroder: I think right now they already look like links, except you can't click them. I think actually making it a link with no other design changes will be intuitive for users. I have clicked on these many times thinking it was a link and been confused that it wasn't. On the candidates page all the cards and photos are clickable so we have already set expectation that these cards should be clickable.

pages/models.py
cjmabry: Agreed about slugs being unreliable but problem with back filling is that we're going to have only partial results be links which is going to be frustrating to user and they're going to think it's a bug. Slug is generated same way it is generated through wagtail so the likelihood of them being different is small, and we can run quick python script to verify they are all the same. We also add step to the add results process which isn't intuitive.

Is there a solution where we don't have to have only some cards on page be links? Perhaps a custom.migration, or perhaps we just verify slug and then make future ticket for scaleable solution?

ericbroder:
@cjmabry I should mention that if we use this approach, then in the long term we can potentially remove candidate field and just leave the candidate endorsement page field instead.

pages/templates/pages/election_tracking_page.html
cjmabry: Afk atm but if I recall correctly only the photos were linked here, correct? At minimum should be entire card similar to candidates page so we're consistent and so it's intuitive for user. We don't necessarily need same hover functionality at the moment but should create issue to make consistent cross-site in future.

cjmabry:
Would like to avoid linking only partial cards if at all possible.

pages/models.py
ericbroder: I suggest we do one of 2 options:
1. make a ticket to update old results to select pages for each of them
2. make slug url the default link if there is no page selected, and page select can just be backup option if slug fails

I suggest 1 because in long term we can replace candidate field with candidate page field, and also I dont understand why candidate model has slug field to begin with since it is not used?

pages/templates/pages/election_tracking_page.html
ericbroder: It's not just photo link, its photo plus name/office/text content - but not results. Similar to candidates page.

I thought about linking results too but it seemed counterintuitive - maybe it should have separate field for linking to actual results article rather than candidate page.

Also the other problem that came up is the social media icons are also links, so if they are enabled we can't have links inside of links.

pages/templates/pages/election_tracking_page.html
```diff
@@ -95,28 +95,45 @@
...
+                                                style="background-image: url({{ candidate.photo.url }})">
+                                                </div>
+                                                {% if candidate_race.candidate_endorsement_page %}
+                                                  </a>
+                                                {% endif %}
+
+                                                <div class="card__content">
+
+                                                  {% if candidate_race.candidate_endorsement_page %}
+                                                    <a href="{% pageurl candidate_race.candidate_endorsement_page %}">
```
ericbroder: @cjmabry think link also covers name/state/office.

pages/models.py
cjmabry: I'm fine with adding a conditional to fallback to slug if there's no candidate page, and then we can make a ticket for updating old results.

We can refactor this process post-election.

ericbroder:
updated to use slug url by default.

pages/templates/pages/election_tracking_page.html
cjmabry: Just reviewed and still only linking photo - are you sure? Don't need to worry about social media links for now.

PR #42, opened 2017-11-07
TECH-406 add results link to candidates page

ericbroder:


pages/templates/pages/candidate_endorsement_index_page.html
```diff
@@ -18,7 +18,10 @@
   </div>
 
   <div class="container pt2 pb2">
-      <h3>2017 Endorsements</h3>
+    <a href="/2017-elections-results/" class="pull-right mt25">Results &raquo;</a>
```
cjmabry: This works, could we add `.fs16` to link class to make it a tad bit bigger and more noticeable? I missed it it first glance but obviously only so much we can do here given timeframe. Good to go after that.

ericbroder:
updated font size

PR #43, opened 2017-11-07
TECH-405 add page link to election result

ericbroder:


ericbroder:
@cjmabry this fixes missing link slug and also if we ignore the commented out social media links then wrapping the whole section in link is better UI.

PR #50, opened 2017-11-16
TECH-290 add BSD authentication client

ericbroder:


bsd/api.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+import os
+from bsdapi.BsdApi import Factory as BsdApiFactory
+from django.conf import settings
+
+
+class Bsd:
+    api = BsdApiFactory().create(
+        id = settings.BSD_API_ID,
```
cjmabry: According to PEP8 (https://www.python.org/dev/peps/pep-0008/) there should be no spaces around around the equals here, but this is an incredibly minor thing. Would just double-check that your IDE is properly linting for PEP8.

bsd/api.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+import os
+from bsdapi.BsdApi import Factory as BsdApiFactory
+from django.conf import settings
+
+
+class Bsd:
```
cjmabry: Technically PEP8 also recommends naming this `BSD`, but up to you because it's a heated debate and I see BSD API doesn't follow this convention apparently. 

https://www.python.org/dev/peps/pep-0008/#id36
```Note: When using abbreviations in CapWords, capitalize all the letters of the abbreviation. Thus HTTPServerError is better than HttpServerError.```

File: bsd/backends.py [hunk 3]
```diff
@@ -0,0 +1,50 @@
...
+
+# TODO: what is best way to do this?
+api = Bsd().api
+
+class BSDAuthenticationBackend:
+
+    def authenticate(self, username=None, password=None):
+
+        # Basic email validation
+        if not '@' in username:
```
cjmabry: According to PEP 8 this should be `if '@' not in username:`.

bsd/backends.py
```diff
@@ -0,0 +1,50 @@
...
+from django.contrib.auth.models import User
+from StringIO import StringIO
+from xml.etree.ElementTree import ElementTree
+from .api import Bsd
+
+
+# TODO: what is best way to do this?
+api = Bsd().api
+
+class BSDAuthenticationBackend:
```
cjmabry: 2 blank lines here, according to PEP8.  https://www.python.org/dev/peps/pep-0008/#id21

bsd/backends.py
```diff
@@ -0,0 +1,50 @@
+from bs4 import BeautifulSoup
```
cjmabry: Can remove since not used.

bsd/backends.py
```diff
@@ -0,0 +1,50 @@
...
+
+        # Basic email validation
+        if not '@' in username:
+            return None
+
+        # Check credentials via BSD api
+        apiResult = api.account_checkCredentials(username, password)
+
+        # Parse and validate response
+        tree = ElementTree().parse( StringIO(apiResult.body) )
```
cjmabry: No whitespace around `StringIO(apiResult.body)` according to PEP 8.

bsd/backends.py
```diff
@@ -0,0 +1,50 @@
...
+            # assertions copied from hydra app
+            assert cons is not None
+            assert cons.find('has_account').text == "1"
+            assert cons.find('is_banned').text == "0"
+
+            # Get user from db that matches email or create new one if needed
+            try:
+                user = User.objects.get(email=username)
+            except User.DoesNotExist:
+                # Create user but dont set password since we use BSD password
```
cjmabry: Should we be creating a user here or is this just a holdover from Hydra? Shouldn't we say "user not found, create an account" or something along those lines? How are we creating a user without a password?

ericbroder: This is not from Hydra. Since we need to return a User model for the authentication backend methods, this seemed like the simplest most standard approach to use. This way we have standard User objects and post-authentication logic can just handle BSD users in normal django fashion by default.

`If no password is provided, set_unusable_password() will be called.`
https://docs.djangoproject.com/en/1.10/ref/contrib/auth/#django.contrib.auth.models.UserManager.create_user
https://docs.djangoproject.com/en/1.10/ref/contrib/auth/#django.contrib.auth.models.User.set_unusable_password

ericbroder: The reason we don't say "user not found, create an account" is because they have already passed BSD api credentials check, and we are just looking for a local user record in db at this point.

In other words, there is a decoupling of bsd authentication step and db user record storage. So the password check happens in bsd but we still have a user record in local db.

bsd/api.py
ericbroder: What do you use to lint for PEP8?

ericbroder:
@cjmabry made some updates and replied to comments, let me know what you think.

bsd/api.py
cjmabry: I use Atom with the Atom IDE integration (https://ide.atom.io/) and ide-python (https://atom.io/packages/ide-python) linter-ui-default (https://atom.io/packages/linter-ui-default).

cjmabry:
Great work!

PR #54, opened 2017-11-20
TECH-529 Build microsite page template.

cjmabry:


pages/models.py
```diff
@@ -62,6 +62,16 @@ class BasePage(Page):
 
 class MicrositePage(Page):
     color_help_text = '6 digit CSS color code.'
+    show_accent_border = models.BooleanField(
+        default=False,
+        help_text='Show solid accent border at top of page.'
+    )
+    accent_border_color = button_background_color = models.CharField(
```
ericbroder: Suggest to sort fields alphabetically.

pages/models.py
```diff
@@ -183,6 +199,9 @@ class MicrositePage(Page):
     twitter_url = models.URLField(null=True, blank=True)
 
     content_panels = Page.content_panels + [
+            FieldPanel('show_accent_border'),
+            FieldPanel('accent_border_color'),
+            FieldPanel('primary_content_embed_code'),
```
ericbroder: I suggest to put the fields at top which are most likely to be edited and accessed regularly. Suggest to put these lower as they are likely to be configured once per page and then probably not needed after that.

pages/static/src/scss/components/_nav.scss
```diff
@@ -561,47 +528,51 @@
...
+      border-bottom: none;
+      display: block;
+    }
+
+    .microsite-header-image {
+      display: block;
+      margin: 50px auto 0;
+    }
+
+    .navbar__social {
```
ericbroder: Suggest to eventually move these out of top left, maybe make follow up ticket? Or add code comments?

pages/static/src/scss/components/_nav.scss
```diff
@@ -561,47 +528,51 @@
...
-//     right: 0;
-// 
-//     > li {
-//       float: none;
-//       display: inline-block;
-//     }
-//   }
-// }
+
+.navbar--microsite {
```
ericbroder: Suggest moving these to microsite scss file? That might be better organization, not sure.

pages/templates/pages/base_page.html
```diff
@@ -91,10 +95,12 @@ <h1>{{ page.title }}</h1>
...
 </div>
 
-{% include "partials/footer.html" %}
-{% include "partials/footer-js.html" %}
+{% block footer %}
+  {% include "partials/footer.html" %}
+  {% include "partials/footer-js.html" %}
 
-{% block footer_scripts %}
+  {% block footer_scripts %}
```
ericbroder: Is nested block pattern correct here? Maybe should be separate blocks? Not sure what footer_scripts is used for.

pages/templates/pages/microsite_page.html
```diff
@@ -1,12 +1,210 @@
 {% extends "pages/base_page.html" %}
 
-{% load wagtailcore_tags %}
+{% load wagtailcore_tags wagtailimages_tags %}
+
+{% block meta %}
+  {{ block.super }}
+  <style>
```
ericbroder: Nice, glad you used style tag. I realized my stub template had inline styles and I was thinking "I hope he doesn't actually do it this way" ;)

pages/templates/pages/microsite_page.html
```diff
@@ -1,12 +1,210 @@
...
+    #custom_header {
+      {% if page.custom_header_background_color %}
```
ericbroder: Instead of if else, could use `firstof` pattern?

pages/templates/pages/microsite_page.html
```diff
@@ -1,12 +1,210 @@
...
+        background-color: #{{page.button_background_color}};
+        border-color: #{{page.button_background_color}};
+      {% endif %}
+
+      {% if page.button_text_color %}
+        color: #{{page.button_text_color}};
+      {% endif %}
+    }
+
+    .microsite-content a#primary-cta {
```
ericbroder: Why styling `#primary-cta` twice? Also don't need selector like this for id tag?

pages/templates/pages/microsite_page.html
```diff
@@ -1,12 +1,210 @@
...
+      {% if page.button_text_color %}
+        color: #{{page.button_text_color}};
+      {% endif %}
+    }
+
+    /* TODO: Add hover color to model or dynamically generate */
+    #primary-cta:hover {
```
ericbroder: Why adding this? Is this not default button behavior?

pages/templates/pages/microsite_page.html
```diff
@@ -1,12 +1,210 @@
...
+      {% if page.primary_content_background_color %}
+        background-color: #{{ page.primary_content_background_color }};
+      {% endif %}
+
+      {% if page.primary_content_text_color %}
+        color: #{{ page.primary_content_text_color }};
+      {% endif %}
+
+      {% if page.primary_content_background_image %}
+        {% image page.primary_content_background_image original as primary_img %}
```
ericbroder: Why need this tag? Image url is not in model already?

pages/templates/pages/microsite_page.html
```diff
@@ -1,12 +1,210 @@
...
+      {% endif %}
+    }
+
+    #secondary_content a {
+      {% if page.secondary_content_text_color %}
+        color: #{{ page.secondary_content_text_color }};
+      {% endif %}
+    }
+
+    #footer, .microsite .footer {
```
ericbroder: Why applying styles to regular footer?

pages/templates/pages/microsite_page.html
```diff
@@ -1,12 +1,210 @@
...
+      {% endif %}
+    }
+
+    #footer a, .microsite .footer a {
+      {% if page.custom_footer_text_color %}
+        color: #{{ page.custom_footer_text_color }};
+      {% endif %}
+    }
+
+    .microsite-content .can_button, .microsite-content #can_embed_form input[type="submit"], .microsite-content #can_embed_form .button, .microsite-content #donate_auto_modal input[type="submit"], .microsite-content #donate_auto_modal .button {
```
ericbroder: Suggest add line breaks to be more readable.

ericbroder: Is there a donate modal on this page?

pages/templates/pages/microsite_page.html
```diff
@@ -1,12 +1,210 @@
...
+        color: #{{page.button_text_color}} !important;
+      {% endif %}
+    }
+  </style>
+{% endblock %}
+
+<div class="microsite">
+
+{% block notification_banner %}
+<!-- TODO: conditional notification banner on page model -->
```
ericbroder: I thought this would follow standard header conditional logic/toggle for now. But maybe this is better? Do we not want to ever show notification banner on a microsite page? Not sure.

pages/templates/pages/microsite_page.html
```diff
@@ -1,12 +1,210 @@
...
+{% endblock %}
 
 {% block content %}
-{% if page.secondary_content_show %}
-<div style="background-color: #{% firstof page.secondary_content_background_color 'FFFFFF' %}">
-	{{ page.secondary_content | richtext }}
-</div>
-{% endif %}
+  <div id="primary_content" class="container-fluid microsite-content">
+    {% if page.custom_header_show %}
```
ericbroder: Custom header is not part of primary content in the model, so this seems confusing?

pages/templates/pages/microsite_page.html
```diff
@@ -1,12 +1,210 @@
...
+              <img src="{{ custom_header_image.url }}" alt="{{page.title}}" class="microsite-header-image">
+            {% endif %}
+
+            {% if page.button_url and page.button_text %}
+              <div class="navbar__buttons">
+                <a id="primary-cta"
+                  href="{{ page.button_url }}"
+                  target="{{ page.button_url_new_window|yesno:'_blank,_self'}}"
+                  class="btn btn-primary"
+                  onclick="trackOutboundLink('{{ page.button_url }}', {{ page.button_url_new_window|yesno:'true,false'}});">
```
ericbroder: Suggest to just leave this enabled always. Field was not really meant for analytics so not sure this logic is desirable.

pages/templates/pages/microsite_page.html
```diff
@@ -1,12 +1,210 @@
...
+        </nav>
+      </div>
+    {% endif %}
+
+    <div class="container pt4 pb4">
+      <div class="col-md-6 mb20 fs18">
+        {{ page.primary_content | richtext }}
+      </div>
+
+      {% if page.primary_content_embed_code %}
```
ericbroder: I thought this would affect layout above as well? Primary content should be more centered if there is no signup form on right side?

pages/templates/pages/microsite_page.html
ericbroder: Remove extra selectors not needed here? I don't see donate modal in page.

pages/models.py
cjmabry: Oh I actually meant to talk about this during review of model but didn't have time - it makes sense to me to put them in the order they go in the page. Going back and forth editing this model was extremely confusing - I essentially had to control F to find anything because they weren't. Fine if I order the rest of them that way - in the order they're presented on the final page?

pages/templates/pages/microsite_page.html
cjmabry: Two reasons I moved away from `firstof` here:
1. It was borking syntax highlighting through entire document for some reason, which was making code editing extremely confusing. I guess because it was in HTML?
2. I didn't want to hard code any values. So for example `firstof text_color FFFFF` would resort to white if no text_color was present, but really we want it to result to whatever color is in the global styles.

pages/templates/pages/microsite_page.html
cjmabry: Need this because `#primary_content a` was overriding this style.

pages/templates/pages/microsite_page.html
cjmabry: No, default behavior is for hover to be a darker color, but this obviously won't work with custom colors, as we'd have to come up with a solution to calculate the dark color.

pages/models.py
ericbroder: I think it's better so that the fields which needs to be maintained at top, and the fields which will only be edited once should be below. Within those groups, I think it makes sense to sort by top down matching layout, which I did. But please do take into account the long term maintenance use case where only really the content fields are updated and the logo is probably not going to change much, for example.

pages/templates/pages/microsite_page.html
cjmabry: It returns an object and there is not a way to access URL without this afaik. This is recommended way to handle images in Wagtail http://docs.wagtail.io/en/v1.9/topics/images.html

pages/templates/pages/microsite_page.html
cjmabry: There are use cases where we would want both nav and custom header I think.

pages/templates/pages/microsite_page.html
cjmabry: This is so you can have a transparent nav with primary background image model behind it if one should choose.

pages/templates/pages/microsite_page.html
cjmabry: What do you mean? This just changes the way we fire the event - if it's opened in new tab we have to do some custom logic which this code is for.

pages/templates/pages/microsite_page.html
cjmabry: Yeah should probably have this be conditional. Will make jira issue.

pages/templates/pages/microsite_page.html
cjmabry: Perhaps not for this AN form, but I just copied AN styles so the button would be changed where it would need to be changed depending on future configurations.

pages/templates/pages/microsite_page.html
cjmabry: This should ideally be a separate field (main footer color and background) but did it quickly because we needed to keep regular footer but change the colors of it to match branding.

pages/templates/pages/base_page.html
cjmabry: This should probably all just go in a footer partial. Footer scripts are javascript that need to go in footer like analytics stuff I believe.

pages/static/src/scss/components/_nav.scss
cjmabry: I thought about that, not sure what made the most sense. Probably should move to microsite.

pages/static/src/scss/components/_nav.scss
cjmabry: I think should have broader conversation since this follows convention of our site nav.

cjmabry:
@ericbroder thanks for review! Agree with most of your points, I will make Jira tickets for them soon but going to merge now so we can get moving.

pages/static/src/scss/components/_nav.scss
ericbroder: Haha I didn't even realize this is same layout as our main site nav 🤦‍♂️

pages/templates/pages/base_page.html
ericbroder: Ok so that was exactly what I was thinking about. We still want analytics even if not showing the main footer right?

pages/templates/pages/microsite_page.html
ericbroder: In this case I suggest to use `!important` instead, or at least add code comments. This looks strange and could easily be changed by accident.

pages/templates/pages/microsite_page.html
ericbroder: I thought we either had standard footer, or custom footer. I did not think we were supporting standard footer but with different colors. If they want to change colors can use the custom footer instead?

pages/templates/pages/microsite_page.html
ericbroder: Ok then suggest to add code comments here, otherwise seems confusing to have selectors that don't appear in markup.

pages/templates/pages/microsite_page.html
ericbroder: Do you mean notification banner instead?

pages/templates/pages/microsite_page.html
ericbroder: I thought the idea was to have the primary background image just for primary content area, not the nav/header area?

pages/templates/pages/microsite_page.html
ericbroder: Suggest only toggle new window based on field, but to trigger analytics events always (or after checking for external url). It's possible we might stay in same window and still go to external url.

pages/templates/pages/microsite_page.html
ericbroder: Ok thanks, fine to do as follow up since current use case will have form.

pages/models.py
cjmabry: I think it's a valid point but I will just say it was very, very confusing creating this this page through the CMS because everything was out of order. A user will always have to create this page, so making that first step confusing to make an optional step (updating the page throughout) that may or may not happen doesn't make the most sense to me. In the event of an update a user would know where to scroll in the page to update said content, whereas if they're not ordered in that way a user will not understand the method by which they are ordered.

pages/templates/pages/base_page.html
cjmabry: Gotcha, yes.

pages/static/src/scss/components/_nav.scss
cjmabry: That was done intentionally and we could potentially A/B test swapping social and donate icons but need a bigger conversation about it.

pages/templates/pages/microsite_page.html
cjmabry: If we want to use regular OR footer with custom colors (which we do in this case) we can't do that without the raw HTML stream field addition because the footer layout is not re-createable 1-to-1 through the WYSIWYG editor. So we either need to add separate fields for OR footer values or add raw html streamfield support.

pages/templates/pages/microsite_page.html
cjmabry: Ha I did not but I realize I misread this. I see them as two distinct elements but perhaps they shouldn't be.

pages/templates/pages/microsite_page.html
cjmabry: The nav should be a part of primary content area so we can enable layouts like these
2017-11-19-21-54-localhost_8000.

pages/templates/pages/microsite_page.html
cjmabry: This does trigger analytics event always, the `true`/`false` just handles the event differently. If set to `true` but not to open in new window then event will still fire. If set to `false` and new window is open event will not properly trigger.

pages/models.py
ericbroder: Either way is fine, not a big deal really.

pages/static/src/scss/components/_nav.scss
ericbroder: I did not mean to swap them. I meant more to rethink what goes in top left area since it's first thing user usually sees.

pages/static/src/scss/components/_nav.scss
ericbroder: I think if possible to keep all them microsite styles together that might be best. I forget if there is a good way to flag dependencies in scss files.

pages/templates/pages/base_page.html
ericbroder: I think this is potentially critical - we want analytics on this page right? Do we need to fix before deploy today?

pages/templates/pages/microsite_page.html
ericbroder: I think it's fine if we want to support standard footer with custom colors, it's just not what I had in mind before so the model is confusing. We may want to clean up a bit at some point to polish this.

ericbroder: 👎  on raw html field 😛

pages/templates/pages/microsite_page.html
ericbroder: Ok I think it's fine to leave as is for now. At some point we can add in conditional toggle for notification banner if we need it.

pages/templates/pages/microsite_page.html
ericbroder: I think idea is fine but not what I had in mind. We may want to update the model.

pages/models.py
ericbroder: @cjmabry I just noticed a typo here, can you take a look at this? i don't even understand why this compiles or works at all, seems like invalid syntax?

PR #60, opened 2017-11-28
Tech 556

cjmabry:


pages/static/src/scss/helpers/_css-helpers.scss
```diff
@@ -1138,6 +1138,22 @@ a.white-text {
   }
 }
 
+.web-store-bg {
```
ericbroder: Suggest we try and have more standard naming conventions in future. SMACSS is pretty good https://smacss.com/

cjmabry: We unofficially use BEM http://getbem.com/ but should solidify and document in near future.

ericbroder: Ok interesting, I'm not familiar with BEM. Yeah let's revisit at some point and get on same page.

ericbroder: @cjmabry made ticket https://ourrevolution.atlassian.net/browse/TECH-558

PR #72, opened 2017-12-13
TECH-564 automate clear cache when group info updated by group leader

cjmabry:


local_groups/models.py
```diff
@@ -56,65 +145,130 @@ class Group(models.Model):
...
-
-    # source = models.CharField(max_length=17, null=True, blank=True)
-    # subsource = models.FloatField(null=True, blank=True)
-    # i_p__address = models.CharField(max_length=16, null=True, blank=True)
-    # constituent__id = models.IntegerField(null=True, blank=True)
-    # signup__id = models.IntegerField(null=True, blank=True)
+    def save(self, *args, **kwargs):
+        # TODO: make main groups url an environment variable
+        # and replace hardcoded /groups throughout site
+        purge_url_from_cache('/groups/' + self.slug)
```
ericbroder: I suggest to clear `/groups` too.

ericbroder: Also is it possible to save first and clear cache after?

local_groups/views.py
```diff
@@ -61,7 +61,7 @@ class GroupUpdateView(LoginRequiredMixin, UpdateView):
 
     # Redirect to same page on success
     def get_success_url(self):
-        return "/groups/" + self.object.slug + "/update/"
+        return "/groups/" + self.object.slug + "/manage/"
```
ericbroder: I replaced this with a named mapping in my PR too. One of us will have to clean up the merge on this. I suggest just revert this line on your end to avoid merge conflict.

ericbroder: https://github.com/Our-Revolution/site/pull/73/files#diff-7984a20a9d14ef5f869b494af09997c1R41

local_groups/models.py
```diff
@@ -56,65 +145,132 @@ class Group(models.Model):
...
+        # and replace hardcoded /groups throughout site
 
-    # source = models.CharField(max_length=17, null=True, blank=True)
-    # subsource = models.FloatField(null=True, blank=True)
-    # i_p__address = models.CharField(max_length=16, null=True, blank=True)
-    # constituent__id = models.IntegerField(null=True, blank=True)
-    # signup__id = models.IntegerField(null=True, blank=True)
+        purge_url_from_cache('/groups')
+        purge_url_from_cache('/groups/' + self.slug)
+        super(Group, self).save(*args, **kwargs)
```
ericbroder: according to docs should be able to clear cache after save:
```

    def save(self, *args, **kwargs):
        do_something()
        super(Blog, self).save(*args, **kwargs) # Call the "real" save() method.
        do_something_else()
```
https://docs.djangoproject.com/en/1.10/topics/db/models/#overriding-predefined-model-methods

PR #80, opened 2017-12-18
TECH-564 automate clear cache when group info updated by group leader

cjmabry:


local_groups/models.py
```diff
@@ -44,7 +44,7 @@ class Group(models.Model):
     group_contact_email = models.EmailField(
         blank=True,
         help_text="""Optional Group Contact Email to publicly display an email
-        different from Individual Contact Email""",
+        different from Group Leader Email""",
```
ericbroder: Should be migration for this?

ericbroder: Should be migration for this?

local_groups/models.py
```diff
@@ -259,18 +259,20 @@ class Group(models.Model):
...
         # and replace hardcoded /groups throughout site
-        super(Group, self).save(*args, **kwargs)
 
-        purge_url_from_cache('/groups/')
-        purge_url_from_cache('/groups/' + self.slug + '/')
+        if self.slug:
+            purge_url_from_cache('/groups')
+            purge_url_from_cache('/groups/' + self.slug)
+
+        super(Group, self).save(*args, **kwargs)
```
ericbroder: Can we save first and clear cache after?

bsd/migrations/0001_initial.py
```diff
@@ -0,0 +1,26 @@
...
+
+    initial = True
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='BSDProfile',
```
ericbroder: why are there bsd migrations in here?

PR #94, opened 2018-01-04
TECH-638 fix caching issues

cjmabry:


varnish.vcl
```diff
@@ -26,17 +26,23 @@ sub vcl_recv {
...
-    if (req.url !~ "^/admin/" && req.url !~ "^/cms/" && req.url !~ "^/docs/" && req.url !~ "^/groups/new/" && req.url !~ "^/organizing-hub/" && req.url !~ "^/join-us-on-slack" && req.url !~ "^/groups/nominations/"){
+    # remove cookie and cache control on every request but ones that start with the following urls
+    # this fixes CSRF issues
+    if (req.url !~ "^/admin" && req.url !~ "^/cms" && req.url !~ "^/docs" && req.url !~ "^/groups/new" && req.url !~ "^/organizing-hub" && req.url !~ "^/join-us-on-slack" && req.url !~ "^/groups/nominations"){
         unset req.http.Cookie;
         unset req.http.Cache-Control;
     }
+    
+    # don't cache the following 
+    if( req.url ~ "^/admin" || req.url ~ "^/cms" || req.url ~ "^/docs" || req.url ~ "^/groups/nominations" || req.url ~ "^/groups/new" || req.url ~ "^/organizing-hub" || req.url ~ "^/join-us-on-slack" ) {
```
ericbroder: Do we really need to repeat the urls again? Can we use `else` instead?

cjmabry: The only reason I'm doing it this way is because it's easy to copy the VCL straight from Fastly, which will minimize errors, should we have to edit this again before we get around to refactoring

ericbroder: I suggest adding that explanation to code comment.

cjmabry: Done.

PR #95, opened 2018-01-05
TECH-677 update links to bsd

ericbroder:


ourrevolution/settings.py
```diff
@@ -180,7 +180,9 @@
 DATABASES = {'default': dj_database_url.config()}
 DATABASES['default']['ENGINE'] = 'django.contrib.gis.db.backends.postgis'
 
+BSD_CREATE_ACCOUNT_URL = 'https://ourrevdev.cp.bsd.net/ctl/Constituent/Login' if os.environ.get('env', 'development') != 'production' else 'https://go.ourrevolution.com/ctl/Constituent/Login'
```
cjmabry: Should this be in .env or do you recommend keeping in settings?

ericbroder: @cjmabry I put into `settings` rather than `.env` based on assumptions:
* doesnt need to be secret, so ok to put into codebase/github (i hope this is true?)
* doesn't really need to be configured per environment given we have only prod and non prod versions, and its unlikely to change anytime soon. so treating more like a constant than an environment config.

ericbroder: which do you think is better?

cjmabry: This is fine, I have no strong preference.

PR #98, opened 2018-01-10
TECH-643 add link to docs in groups portal

cjmabry:


pages/templates/pages/group_resource_page.html
```diff
@@ -37,8 +39,12 @@ <h3 class="panel-title">Contents</h3>
 <script>
   $(document).ready(function() {
 
+    // Get if user is logged in
+    var $is_authenticated = {{ user.is_authenticated|yesno:"true,false" }}
```
ericbroder: missing semicolon at end of line?

pages/templates/pages/group_resource_page.html
```diff
@@ -58,10 +64,14 @@ <h3 class="panel-title">Contents</h3>
...
 
-      // Update layout and display table of contents
-      $contentBody.removeClass('col-md-offset-3');
       $contentToc.removeClass('hidden');
     }
+
+    if ($headings.length || $is_authenticated) {
+      $contentSidebar.removeClass('hidden');
+      $contentBody.addClass('col-md-9');
+      $contentBody.removeClass('col-md-8 col-md-offset-2');
```
ericbroder: This logic for layout is strange to me. If there is no sidebar, then main content area is 8 columns. And if there is a sidebar, then main content is 9 columns. So that means the main content is larger area when there is also a sidebar? I would think main content area should be either the same or smaller when sidebar is added? Or in other words, when sidebar is absent, main content area should be same or larger?

cjmabry: The only reason it is 8 columns is because you can't center 9 columns within bootstrap, and I didn't want to rewrite the whole layouts just for this one scenario. Without sidebar content should be centered, so had to go to 8 there.

local_groups/templates/partials/group_portal_nav.html
```diff
@@ -16,5 +16,10 @@
   href="{{ organizing_guides_url }}">
   Organizing Guides
   </a>
-
+  
+  {% with link_text="Documentation" link_url="/docs/"  %}
```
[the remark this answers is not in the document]
cjmabry: Good call.

pages/templates/pages/group_resource_page.html
```diff
@@ -1,11 +1,9 @@
 {% extends "group_portal.html" %}
 
-{% load bootstrap3 wagtailcore_tags %}
+{% load bootstrap3 wagtailcore_tags local_groups_tags %}
 
 {% block nav %}
```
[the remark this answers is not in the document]
cjmabry: We need this to be able to use:
```
{% group_portal_nav %}
```

pages/templates/pages/group_resource_page.html
```diff
@@ -14,8 +12,12 @@
 
   <div class="row pt4 pb4">
 
-    <div class="col-md-3 hidden" id="content-body-toc">
-      <div class="panel panel-default">
+    <div class="col-md-3 hidden" id="content-sidebar">
+      {% if user.is_authenticated %}
```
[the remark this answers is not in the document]
cjmabry: Yup! https://docs.djangoproject.com/en/dev/topics/auth/default/#authentication-data-in-templates

pages/templates/pages/group_resource_page.html
cjmabry: Thanks for catching.

cjmabry:
@ericbroder Thanks for feedback, changes made.

local_groups/templates/partials/group_portal_nav.html
```diff
@@ -17,4 +17,7 @@
   Organizing Guides
   </a>
 
+  <a class="list-group-item {% if organizing_docs_url == request.path %}active{% endif %}" href="{{ organizing_docs_url }}">
```
ericbroder: nice 👍

local_groups/templatetags/local_groups_tags.py
```diff
@@ -26,5 +26,6 @@ def group_portal_nav(context):
     return {
         'group': group,
         'organizing_guides_url': settings.ORGANIZING_GUIDES_URL,
+        'organizing_docs_url': settings.ORGANIZING_DOCS_URL,
```
ericbroder: 👍

ourrevolution/settings.py
```diff
@@ -247,6 +247,7 @@
 MEDIA_URL = '/media/' if os.environ.get('env', 'development') != 'production' else 'https://s3.amazonaws.com/our-revolution-dot-com/'
 
 ORGANIZING_GUIDES_URL = '/docs/organizing-guides/'
+ORGANIZING_DOCS_URL = '/docs'
```
ericbroder: Mostly good, but I suggest to add trailing slash, and suggest alpha sort the variables here (docs above guides).

pages/templates/pages/group_resource_page.html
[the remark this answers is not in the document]
ericbroder: Are you sure about that? I don't understand why. `group_portal_nav` actually goes into the content region, not the header nav area. There's no logic here other than outputting the `super` version, so what's the point?

pages/templates/pages/group_resource_page.html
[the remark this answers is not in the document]
ericbroder: Good to know.

pages/templates/pages/group_resource_page.html
ericbroder: It's relatively minor but still seems strange to me.

ourrevolution/settings.py
ericbroder: @cjmabry actually this breaks logic without a trailing slash. When current page is docs page, menu item not showing up as active.

local_groups/templates/group_dashboard.html
```diff
@@ -122,7 +122,7 @@ <h3 class="mb20 mt0">
           {% endcomment %}
 
           <p>
-            <strong>Need help?</strong> You can reach us at
+            <strong>Need help?</strong> Visit our <strong><a href="/docs">documentation</a></strong>, or reach us at
```
ericbroder: Let's use the setting here too?

PR #112, opened 2018-01-31
TECH-655 TECH-707 TECH-487 move candidate fields to page model

ericbroder:


pages/models.py
```diff
@@ -548,64 +548,197 @@ def get_context(self, *args, **kwargs):
...
+    facebook_url = models.URLField(null=True, blank=True)
+    general_election_date = models.DateField(null=True, blank=True)
+    general_election_result = models.CharField(
+        max_length=result_max_length,
+        choices=result_choices,
+        null=True,
+        blank=True
+    )
+    instagram_url = models.URLField(null=True, blank=True)
+    office = models.CharField(null=True, blank=True, max_length=128)
```
cjmabry: Should this be required?

pages/models.py
```diff
@@ -548,64 +548,197 @@ def get_context(self, *args, **kwargs):
...
+    photo = models.ForeignKey(
```
cjmabry: Should this be required?

pages/models.py
```diff
@@ -548,64 +548,197 @@ def get_context(self, *args, **kwargs):
...
+        blank=True
+    )
+    social_image = models.ForeignKey(
+        'wagtailimages.Image',
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name='+'
+    )
+    state_or_territory = models.CharField(null=True, blank=True, max_length=32)
```
cjmabry: Can we change this to `USStateField` to avoid retyping states/territories and eliminate spelling errors? We are already using localflavor. https://github.com/myusuf3/django-localflavor-us/blob/master/django_localflavor_us/us_states.py

pages/models.py
```diff
@@ -548,64 +548,197 @@ def get_context(self, *args, **kwargs):
...
+    youtube_url = models.URLField(null=True, blank=True)
 
+    parent_page_types = ['pages.CandidateEndorsementIndexPage']
 
     content_panels = Page.content_panels + [
-            FieldPanel('body', classname="full"),
-            FieldPanel('candidate'),
-            FieldPanel('election')
-        ]
+        MultiFieldPanel(
```
cjmabry: I really like the `MultiFieldPanel` usage here and think this is really intuitive UX to group them in this way. No action needed here.

pages/models.py
```diff
@@ -548,64 +548,197 @@ def get_context(self, *args, **kwargs):
...
         'endorsements.Election',
         null=True,
         blank=True,
         on_delete=models.SET_NULL
     )
-    signup_tagline = models.CharField(max_length=128, blank=True, null=True)
-    social_image = models.ForeignKey('wagtailimages.Image', null=True, blank=True, on_delete=models.SET_NULL, related_name='+')
-    parent_page_types = ['pages.CandidateEndorsementIndexPage']
+    facebook_url = models.URLField(null=True, blank=True)
+    general_election_date = models.DateField(null=True, blank=True)
```
cjmabry: Should this be required? Are there scenarios where someone wouldn't have a general election?

pages/models.py
```diff
@@ -548,64 +548,197 @@ def get_context(self, *args, **kwargs):
...
+            return self.primary_election_date
+        # Return primary date if candidate lost primary
+        elif (
+            self.general_election_result is None and
+            self.primary_election_result == 'loss'
+        ):
+            return self.primary_election_date
+        # Return general date otherwise
+        else:
+            return self.general_election_date
```
cjmabry: This can fail if `general_election_date` is not required. Either need to make `general_election_date` required or provide fallback here.

pages/templates/pages/candidate_endorsement_index_page.html
```diff
@@ -55,20 +55,54 @@ <h3 class="pull-left mb0">
...
-                              <span class="card__item name">{{ candidate.name }}</span>
-                              <span class="card__item bold smaller state">{{ candidate.state }} {{ candidate.district }}</span>
-                              <span class="card__item bottom">{% if candidate.primary_date %}<div class="card__item smaller bold">Election: <span class="date hidden">{{ candidate.primary_date|date:"U" }}</span>{{ candidate.primary_date }}</div> {% endif %}{{ candidate.office }}</span>
-                          </div>
-                      </a>
-                  </li>
-              {% endwith %}
+
+          {% for candidate in candidates %}
+          {% with candidateendorsementpage=candidate.candidateendorsementpage %}
```
cjmabry: Since we don't have established style guidelines for coding Django templates there is no action needed here, but shouldn't this be indented? It would make the compiled source code uglier but would make the template more readable. Curious to hear your thoughts here.

pages/templates/pages/candidate_endorsement_index_page.html
```diff
@@ -55,20 +55,54 @@ <h3 class="pull-left mb0">
...
+
+          <li class="card clearfix">
+
+            <a href="{% pageurl candidateendorsementpage %}">
+
+              {% image candidateendorsementpage.photo original as photo_img %}
```
cjmabry: What's reasoning behind `candidateendorsementpage` instead of `candidate_endorsement_page`? Probably need to establish guidelines here, too, but latter seems more readable.

pages/templates/partials/sign-up.html
```diff
@@ -2,7 +2,13 @@
     {% if not page.issuepage %}<div class="col-md-6 col-md-offset-3">{% endif %}
         <h3>
             {% if page.candidateendorsementpage %}
-                Join the fight to get {{ page.candidate.name }} elected.
+                Join the fight to get
+                {% if page.candidate %}
+                {{ page.candidate.name }}
+                {% else %}
+                {{ page.title }}
```
cjmabry: If we're going to treat title as name needs to be clear to an admin when creating page. Can we add/override help-text on title field? Don't want `Join the fight to get Our Revolution Endorses Mike Connolly elected` by accident if admins misinterpret page title.

pages/models.py
ericbroder: Not required because:
- we are supporting legacy records with these fields empty (most important)
- user can save draft in progress with some fields empty
- if candidate loses in primary we don't really need the general election date

pages/models.py
ericbroder: ditto

pages/models.py
ericbroder: ditto

pages/models.py
ericbroder: @cjmabry My concern about this is that the state field does not have clear predefined values to choose from. I renamed it to state_or_territory field to be more accurate/flexible, because Puerto Rico and DC are not states. Although this does not even make sense always. When Keith Ellison ran for DNC Chair we listed state as Minnesota, but really it's Location = DNC or United States? Or what about president, should be Location = United States? Or are we going to just list the candidate's home state?

I'm curious to hear your thoughts on this, but this is why I avoided dropdown menu.

pages/models.py
ericbroder: It will return None as a fallback, I think that's fine? Since the fields are optional then this function also returns an optional value basically.

pages/templates/pages/candidate_endorsement_index_page.html
ericbroder: I like when the IDE does the indenting standard for me, and in this case the IDE says this is correct. I'm not sure how to verify the correct standard here since like you said we don't have official guidelines.

pages/templates/pages/candidate_endorsement_index_page.html
ericbroder: Actually I'm a bit confused about the model here, but `candidate.candidateendorsementpage` is something that already exists for some reason. So all I did was shorten it and remove the prefix basically. I'm confused why the page object is somehow nested one level down, since the logic seems like it should be returning a list of page objects?

pages/templates/partials/sign-up.html
ericbroder: @cjmabry it's a good idea but I'm not sure how to do this. See: https://github.com/wagtail/wagtail/issues/161

ericbroder:
@cjmabry please see replies.

pages/models.py
cjmabry: 👍

pages/templates/pages/candidate_endorsement_index_page.html
cjmabry: I actually read a bit about this and most IDE's don't care because there aren't actual official style guides here like there are for regular Python with PEP 8. Just something to think about and maybe standardize in future.

pages/templates/partials/sign-up.html
cjmabry: Ok fine for now if we can't do it, let's remember to specify this in any documentation.

ericbroder: Yeah I agree should be called out in our training/documentation.

pages/models.py
cjmabry: This should be optional in the case of Ellison's DNS run or president. `USStateField` includes all territories including DC and PR. I think calling state or territory is fine, but the idea is to remove a step of typing state names and eliminate mistakes by switching to predefined elections. This will also help standardize data (this was something David said was really helpful on candidate endorsement applications and might help us auto-populate data here as well). If there's a scenario where state or territory isn't applicable (DNC chair, president) we can leave blank.

In the president's case we should probably add a country field but outside of scope of this ticket.

cjmabry:
@ericbroder need reply on `USStateField` stuff.

pages/models.py
ericbroder: @cjmabry I'm ok with trying to use `USStateField` here. But I don't think it should be blank for edge cases, probably better to just put candidate home state as we have been for DNC/POTUS etc.

ericbroder:
@cjmabry I added `USStateField` 👍

PR #116, opened 2018-02-06
TECH-286 add member newsletter to cms

ericbroder:


pages/templates/partials/candidate_endorsement.html
```diff
@@ -128,7 +128,7 @@ <h1>{{ page.title }}</h1>
               </div>
               <div class="col-md-6 col-sm-9 col-xs-8">
                   <div class="candidate-bio">
-                      {{ page.body | safe }}
+                      {{ page.body | richtext }}
```
cjmabry: What's this?

cjmabry:
Great work, hold on merging until we deploy.

pages/templates/partials/candidate_endorsement.html
ericbroder: I think there's a bug where some rich text fields are using the wrong filter. `safe` is a django filter for html fields. `richtext` is a wagtail filter specifically for rich text fields. Does that make sense? This is a rich text field so that's why I changed it.

PR #117, opened 2018-02-07
TECH-790 Distinguish state groups on the map

cjmabry:


local_groups/models.py
```diff
@@ -33,6 +33,21 @@ class Group(models.Model):
         blank=False,
         unique=True
     )
+
+    GROUP_TYPES = (
+        ('state-organizing-committee', 'State Organizing Committee'),
+        ('state-chapter', 'State Chapter'),
+        ('campus', 'Campus'),
+        ('local-group', 'Local Group')
```
ericbroder: I generally suggest to use integer IDs for stuff like this. The point of the key is that it should be a unique ID but doesn't need to be human readable. Plus then if we change the display name we don't have to worry about keeping the key in sync.

local_groups/migrations/0061_auto_20180207_0131.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('local_groups', '0060_auto_20171213_2208'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
ericbroder: are you sure this is safe? going to wipe out existing data?

local_groups/models.py
```diff
@@ -203,8 +218,6 @@ class Group(models.Model):
         max_length=250
     )
 
-    state_organizing_committee = models.BooleanField(default=False)
```
ericbroder: Please double check this is safe. Also please be careful about A/B deploy issues. Usually before removing a field from db need to remove it from code in first deploy, then drop from db in later deploy.

pages/models.py
```diff
@@ -1272,7 +1272,7 @@ def get_context(self, *args, **kwargs):
 class GroupPage(RoutablePageMixin, Page):
     @route(r'^$')
     def index_view(self, request):
-        groups = Group.objects.all().order_by('state_organizing_committee')
+        groups = Group.objects.all().order_by('group_type')
```
ericbroder: What is the purpose of the sorting here?

pages/static/src/js/groups.js
```diff
@@ -83,11 +115,18 @@ module.exports = function() {
...
+        // if SOC, color red and put on top layer
+        marker = L.marker(coords.reverse(), {icon: redIcon, zIndexOffset: 1000})
+      } else {
+        marker = L.marker(coords.reverse(), {icon: blueIcon});
+      }
+      
       marker.properties = group.properties;
 
-      marker.bindPopup('<h4>'+ group.properties.name + '</h1>');
+      marker.bindPopup('<h4>'+ group.properties.name + '</h1><a href="/groups/'+ group.properties.slug +'" class="component__cta btn btn-block btn-primary uppercase ls2">Get Involved</a>');
```
ericbroder: We should ideally put `/groups/` into settings config if we haven't already. Also suggest to put trailing slash in url.

pages/static/src/js/groups.js
```diff
@@ -104,7 +143,7 @@ module.exports = function() {
   }
 
   function updateInfo(groups) {
-    var location = '', city, county, state, country;
+    var location = '', city, county, state, country, group_type = null;
```
ericbroder: Why need to add `= null`?

pages/static/src/js/groups.js
```diff
@@ -141,12 +180,23 @@ module.exports = function() {
 
         // TODO: show recurring meeting here instead
         description = groups[i].properties.description;
+        group_type = groups[i].properties.group_type;
+        
+        if (group_type == 'state-organizing-committee') {
+          group_type = 'State Organizing Committee';
```
ericbroder: Can we access the model we made for this? Otherwise seems not very DRY to write out display name again here.

pages/static/src/js/groups.js
```diff
@@ -141,12 +180,23 @@ module.exports = function() {
...
+        } else {
+          group_type = null;
+        }
 
         $('.groups-map-info__status').append('\
           <div class="component">\
             <div class="component__heading">\
-              <span class="component__location">' + location + '</span>\
-              <h4 class="component__name">' + groups[i].properties.name + ((groups[i].properties.state_organizing_committee) ? " (SOC) " : "") + '</h4>\
+              <span class="component__location">' + location + ((group_type) ? ' - ' + group_type : "") + '</span>\
```
ericbroder: Is it correct that you are adding this next to location rather than name?

pages/templates/pages/group_index_page.html
```diff
@@ -59,7 +34,7 @@ <h4 class="mb20 mt0">No group in your area?</h4>
             <a href="https://docs.google.com/document/d/1BWp6HCZ6tngr6SJHJB3H1uPTX2Hcv6cMUQondCrBkHg/edit" class="btn btn-block btn-primary uppercase ls2" target="_blank" onclick="trackOutboundLink('https://docs.google.com/document/d/1BWp6HCZ6tngr6SJHJB3H1uPTX2Hcv6cMUQondCrBkHg/edit', true)">Start a group</a>
             <a href="/groups/new" class="btn btn-block btn-secondary uppercase ls2">Add your group</a>
 
-            <p class="mt20"><small>We're currently experiencing a high number of group applications, which may lead to delays in adding your group to the map.</small></p>
+            <p class="mt0"><small><div class="groups-map__icon red small"></div> State Organizing Committee or State Chapter</small></p>
```
ericbroder: nice 👍

local_groups/models.py
```diff
@@ -33,6 +33,21 @@ class Group(models.Model):
...
         unique=True
     )
+
+    GROUP_TYPES = (
+        ('state-organizing-committee', 'State Organizing Committee'),
+        ('state-chapter', 'State Chapter'),
+        ('campus', 'Campus'),
+        ('local-group', 'Local Group')
+    )
+    group_type = models.CharField(
```
ericbroder: Should this field be required? Not sure why it's optional.

ericbroder: Also how are we going to populate the correct values in prod?

local_groups/migrations/0061_auto_20180207_0131.py
cjmabry: We only have four groups marked, and we're manually remarking them anyways, so wanted to avoid having to do separate deploys due to time-sensitivity.

local_groups/models.py
cjmabry: Agreed, except we need these values in the exported CSV so used strings here, otherwise we get group type 1, 2, 3, or 4 without changing export code.

local_groups/models.py
cjmabry: This was never used outside of admin/really at all. We're manually setting group status after deploy.

pages/models.py
cjmabry: So when results are returned in sidebar they are sorted by priority (state organizing committee first).

pages/static/src/js/groups.js
cjmabry: 👍 leftover from different implementation will remove.

pages/static/src/js/groups.js
cjmabry: Methods like `get_foo_display` aren't available with current implementation, otherwise agree.

local_groups/models.py
cjmabry: Will populate manually, initially wasn't going to make required because if someone submitted a new group it would bork things, but working on this now.

cjmabry: Is now required.

local_groups/models.py
ericbroder: "we need these values in the exported CSV" - what do you mean by this? why do we need it?

local_groups/migrations/0061_auto_20180207_0131.py
ericbroder: I suggest let's remove it from app logic in one deploy, and drop from database in later deploy, to avoid downtime. A/B deploy issue and preserving legacy data issue are two separate things. If we don't need the legacy data that's fine, but we still should avoid throwing errors during deploy.

local_groups/models.py
ericbroder: Sounds fine to delete legacy data, but still should try to avoid downtime during deploys.

pages/models.py
ericbroder: But isn't it just going to sort alphabetically? I don't see where the group types are ranked in order of which should appear first. This looks like a string alpha sort?

pages/static/src/js/groups.js
ericbroder: Hmm, I'm not sure how to access the Model from a js file. I guess something to think about/research for future.

pages/static/src/js/groups.js
ericbroder: @cjmabry just want to double check this is right?

local_groups/models.py
```diff
@@ -33,6 +33,21 @@ class Group(models.Model):
...
+        max_length=64,
+        blank=False,
+        null=False,
+        choices=GROUP_TYPES,
+        default='local-group'
```
ericbroder: nice 👍

local_groups/models.py
```diff
@@ -33,6 +33,21 @@ class Group(models.Model):
...
+    GROUP_TYPES = (
+        ('state-organizing-committee', 'State Organizing Committee'),
+        ('state-chapter', 'State Chapter'),
+        ('campus', 'Campus'),
+        ('local-group', 'Local Group')
+    )
+    group_type = models.CharField(
+        max_length=64,
+        blank=False,
+        null=False,
```
ericbroder: I think required is default? Can leave out blank/null params.

pages/models.py
```diff
@@ -1306,6 +1306,7 @@ def add_group_view(self, request):
 
                 # Get new group id
                 group.group_id = str(self.get_new_group_id())
+                group.group_type = 'local-group'
```
ericbroder: Why do we need this is we already have a default set in the model for this field?

pages/templates/pages/add_group.html
```diff
@@ -224,7 +224,8 @@ <h4 class="mb0">{% bootstrap_label "Regular Meeting" label_for="recurring_meetin
           </div>
         </div>
 
-      <input type="hidden" id="id_group_id" name="group_id" value='9999'>
+        <input type="hidden" id="id_group_id" name="group_id" value='9999'>
+        <input type="hidden" id="id_group_type" name="group_type" value='local-group'>
```
ericbroder: I suggest we remove these hidden fields. This is kind of an anti-pattern - aren't we basically trying to say these fields don't belong on this form because the user is not really supplying these values? We are going to set these values later so they should not appear in the markup. Also users can edit markup in browser before submitting, so this is not a reliable approach.

local_groups/models.py
cjmabry: Organizing team exports CSVs of groups for various organizing purposes.

ericbroder: Ok sure. I suggest adding code comment here to explain this, it's not really intuitive that these need to be semantic human readable keys for this reason. Also we could potentially use integer IDs and document what the id numbers correspond to. Or maybe we can insert the display name into csv, or potentially both the key and the display name.

pages/templates/pages/add_group.html
cjmabry: This is a required value, so unless we set blank = true, this is required on all forms. If we do set blank = true, we'll get an error on admin side if a group_type is saved as none because null = false. If we set null = false, we'll allow null in these fields. So this was only solution that made sense.

cjmabry: Disregard found a solution.

ericbroder: I think there is a way to separate the form logic from the model logic. The model stored in db has certain fields that are required, but if they are generated automatically they shouldn't be on the form. For example `ID` and `date_created` fields are usually generated automatically, and don't appear on a form. So there needs to be some separation of logic in some cases between forms and models. I don't think putting them as hidden input fields in html with invalid data values is a good solution, and I suggest we avoid this.

ericbroder: I understand we have a lot of legacy code and tech debt etc., but still want to establish what is our standard here, and I think it should be decoupling form logic from model logic when needed in cases like this.

pages/static/src/js/groups.js
cjmabry: Is fine, classname should probably not be called location but is generally just a subheading.

PR #118, opened 2018-02-07
TECH-790 Distinguish state groups on the map

cjmabry:


local_groups/forms.py
```diff
@@ -50,7 +50,9 @@ class GroupCreateForm(forms.ModelForm):
 
     class Meta:
         model = Group
-        exclude = ('slug', 'signup_date', 'status', 'point')
+        exclude = (
+            'slug', 'signup_date', 'status', 'point', 'group_id', 'group_type'
```
ericbroder: nice 👍

ericbroder: I wonder if need to update any other forms too? Manage Group form?

ericbroder: I wonder if would be better to have a whitelist of fields rather than exclude list? For case like this new field would be excluded automatically if we had field whitelist.

local_groups/models.py
```diff
@@ -34,20 +34,23 @@ class Group(models.Model):
         unique=True
     )
 
+    # Order by group priority
```
ericbroder: 👍

local_groups/models.py
```diff
@@ -34,20 +34,23 @@ class Group(models.Model):
...
     GROUP_TYPES = (
-        ('state-organizing-committee', 'State Organizing Committee'),
-        ('state-chapter', 'State Chapter'),
-        ('campus', 'Campus'),
-        ('local-group', 'Local Group')
+        (1, 'State Organizing Committee'),
```
ericbroder: 👍

local_groups/models.py
```diff
@@ -34,20 +34,23 @@ class Group(models.Model):
...
+        (2, 'State Chapter'),
+        (3, 'Campus'),
+        (4, 'Local Group')
     )
-    group_type = models.CharField(
-        max_length=64,
+    group_type = models.IntegerField(
```
ericbroder: 👍

local_groups/models.py
```diff
@@ -34,20 +34,23 @@ class Group(models.Model):
...
         blank=False,
         null=False,
         choices=GROUP_TYPES,
```
ericbroder: Does `choices` work for integer field?

local_groups/models.py
```diff
@@ -34,20 +34,23 @@ class Group(models.Model):
...
-        default='local-group'
+        default=4
     )
 
+    # TODO: Remove after deploy TECH-794
```
ericbroder: 👍

pages/models.py
```diff
@@ -1272,7 +1272,8 @@ def get_context(self, *args, **kwargs):
 class GroupPage(RoutablePageMixin, Page):
     @route(r'^$')
     def index_view(self, request):
-        groups = Group.objects.all().order_by('group_type')
+        # Order is set by integer value in GROUP_TYPES tuple in models
```
ericbroder: 👍

pages/models.py
```diff
@@ -1272,7 +1272,8 @@ def get_context(self, *args, **kwargs):
...
+        groups = Group.objects.all().order_by('-group_type')
```
ericbroder: Why add minus sign? Was it wrong before?

pages/models.py
```diff
@@ -1306,7 +1307,6 @@ def add_group_view(self, request):
 
                 # Get new group id
                 group.group_id = str(self.get_new_group_id())
-                group.group_type = 'local-group'
```
ericbroder: 👍

pages/static/src/js/groups.js
```diff
@@ -181,13 +182,14 @@ module.exports = function() {
         // TODO: show recurring meeting here instead
         description = groups[i].properties.description;
         group_type = groups[i].properties.group_type;
-        
-        if (group_type == 'state-organizing-committee') {
+
+        // See GROUP_TYPES tuple in local groups model for group types
```
ericbroder: 👍

pages/templates/pages/add_group.html
```diff
@@ -224,8 +224,6 @@ <h4 class="mb0">{% bootstrap_label "Regular Meeting" label_for="recurring_meetin
           </div>
         </div>
 
-        <input type="hidden" id="id_group_id" name="group_id" value='9999'>
-        <input type="hidden" id="id_group_type" name="group_type" value='local-group'>
```
ericbroder: 👍

pages/templates/pages/group_page.html
```diff
@@ -15,7 +15,7 @@
 
       <div class="container pt6">
         <p class="mb0">
-          {% if group.group_type != 'state-organizing-committee' %}
+          {% if group.group_type != 1 %}
```
ericbroder: 👍  I suggest adding code comment too for this section.

local_groups/models.py
cjmabry: Yup!

pages/models.py
cjmabry: Default sort was descending for whatever reason.

pages/templates/pages/group_page.html
ericbroder: Thanks, although I meant more like "Show different content for SOC vs whatever". More like to explain the logic in general here.

PR #125, opened 2018-02-20
TECH-785 More visibility on why an application is incomplete

cjmabry:


nominations/models.py
```diff
@@ -355,6 +470,76 @@ def save(self, *args, **kwargs):
...
+            if self.questionnaire.status == 'incomplete':
+                # needs questionaire
+                status = 'needs-questionnaire'
+            else:
+                # questionnaire complete
+                status = 'incomplete'
+        return status
+
+    def is_editable(self):
+        """Returns whether the user can edit this application."""
```
ericbroder: What does it mean `user can edit this application`? Seems like data can be edited outside of these statuses. Which user are we talking about?

nominations/models.py
```diff
@@ -355,6 +470,76 @@ def save(self, *args, **kwargs):
...
+            'needs-staff-review',
+            'under-review',
+            'approved',
+            'removed',
+            'expired',
+            'hold'
+        ]
+
+        if self.status not in exempt_statuses:
+            self.status = self.generate_application_status()
```
ericbroder: I suggest moving this "exempt" logic to `generate_application_status` method. Then this part here can be rewritten like
```
status_new = self.generate_application_status()
if self.status != status_new:
            self.status = status_new
```
or simply
```
self.status = self.generate_application_status()
```

nominations/models.py
```diff
@@ -355,6 +470,76 @@ def save(self, *args, **kwargs):
...
+
+    def save(self, *args, **kwargs):
+        if not self.nomination or not self.questionnaire:
+            self.create_related_objects()
+
+        self.auto_populate_research_fields()
+
+        # TODO move to settings.py or env?
+        # dont run status generator if set to these
+        exempt_statuses = [
```
ericbroder: Is this just the inverse of editable status list?

Also in case like this I suggest to use whitelist instead of blacklist. If we create a new status and don't update the logic here, what should we expect to be default behavior? Probably would prefer to treat as exempt by default? Although I suggest just move this logic into the status getter method anyway and just set the status every time.

nominations/templates/application_status.html
```diff
@@ -32,20 +32,19 @@ <h1 class="mb0 mt0">{{application}}</h1>
...
             <li class="list-group-item"><a href="/groups/nominations/questionnaire?id={{application.id}}"><span class="glyphicon {% if application.questionnaire.status == 'complete' %}glyphicon-check{% else %}glyphicon-unchecked{% endif %}"></span> 3. Have your candidate fill out the candidate questionaire</a></li>
-            
-            {% if application.status == 'incomplete' %}
-              <li class="list-group-item"><a href="/groups/nominations/submit?id={{application.id}}"><span class="glyphicon {% if application.status == 'submitted' %}glyphicon-check{% else %}glyphicon-unchecked{% endif %}"></span> 4. Submit your nomination</a></li>
-            {% else %}  
+
+            {% if application.is_editable %}
+              <li class="list-group-item"><a href="/groups/nominations/submit?id={{application.id}}"><span class="glyphicon glyphicon-unchecked"></span> 4. Submit your nomination</a></li>
+            {% else %}
               <li class="list-group-item"><span class="glyphicon glyphicon-check"></span> 4. Submit your nomination</li>
```
ericbroder: Suggest to make this more DRY - don't repeat `4. Submit your nomination` twice in the code.

nominations/templates/questionnaire.html
```diff
@@ -30,7 +30,9 @@ <h1 class="mb0 mt0">Candidate Questionnaire</h1>
           {% if not questionnaire.status == 'complete' %}
             <div class="alert alert-danger br3"><strong>Note:</strong> We strongly recommend having the candidate fill out this form because their level of alignment with our national platform is an important factor in our endorsement decision. You can send this form to the candidate <a href="/groups/nominations/questionnaire/?id={{application.id}}">here</a>.</div>
           {% endif %}
-          {% if application.status == 'submitted' or questionnaire.status == 'complete' %}
+          
+          {% comment %} if application isn't editable or questionnaire is complete, display read-only fields to user {% endcomment %}
+          {% if not application.is_editable or questionnaire.status == 'complete' %}
```
ericbroder: Can we just simplify this to `{% if questionnaire.status == 'complete' %}`?

nominations/tests.py
```diff
@@ -1,3 +1,87 @@
 from django.test import TestCase
+from models import Application
+from local_groups.models import Group
 
-# Create your tests here.
+
+class ApplicationTestCase(TestCase):
```
ericbroder: 😂

nominations/tests.py
```diff
@@ -1,3 +1,87 @@
...
+    def setUp(self):
+        # create group
+        group = Group.objects.create(
```
ericbroder: Is this really editing db or is this mocked somehow? For unit test stuff like this should be mocked.

nominations/tests.py
```diff
@@ -1,3 +1,87 @@
...
+            candidate_state='TN'
+        )
+
+    def test_application_creation(self):
+        """
+        When an application is created, a Questionnaire and a Nomination
+        should be created as well.
+        """
+
+        app = Application.objects.first()
```
ericbroder: If this is a unit test, then we should test an individual method here? What are we testing, the save method? This is really written more like an integration test than a unit test right now.

nominations/tests.py
```diff
@@ -1,3 +1,87 @@
...
+
+        # check for nominations and questionnaire
+        self.assertNotEqual(app.nomination, None)
+        self.assertNotEqual(app.questionnaire, None)
+        self.assertEqual(app, app.nomination.application)
+        self.assertEqual(app, app.questionnaire.application_set.first())
+
+    def test_status_generation(self):
```
ericbroder: I suggest to make this a unit test, this test should just focus on the `generate_application_status` method and make sure it returns the correct value based on test set up. Then there should be a separate unit test for `save` method that makes sure the `generate_application_status` method is called at the appropriate times. So this way the **unit** tests are very granular.

nominations/tests.py
```diff
@@ -1,3 +1,87 @@
...
+        self.assertEqual(app.questionnaire.status, 'incomplete')
+        self.assertEqual(app.status, 'needs-questionnaire')
+
+        # complete questionnaire
+        app.questionnaire.status = 'complete'
+        app.questionnaire.save()
+        self.assertEqual(app.nomination.status, 'complete')
+        self.assertEqual(app.questionnaire.status, 'complete')
+        # I don't know why this is failing
+        self.assertEqual(app.status, 'incomplete')
```
ericbroder: I don't understand what `application_set` is or how this logic is supposed to work.

ericbroder:
@cjmabry In trying to understand the editable/exempt status logic, it seems to me that perhaps what we are trying to do is group the statuses by which are pre-submitted ("editable" statuses?) and which are post-submitted statuses ("exempt" statuses?). If that is true, then I suggest the following pattern:
- make a list of status choices called `statuses_unsubmitted` or something like that which contains the pre-submitted/editable statuses
- make a list of status choices called `statuses_submitted` or something like that which contains the post-submitted/exempt statuses
- make a master list that combined the two above lists in a DRY fashion which does not duplicate the status definitions, and use this combined list as the choices value for the schema
- make a method called `is_submitted` or something like that which checks if the current status belongs in the pre-submitted or the post-submitted lists

nominations/models.py
```diff
@@ -179,57 +192,170 @@ class Application(models.Model):
...
+            'needs-group-form-and-questionnaire',
+            'Needs Group Form and Questionnaire'
+        ),
+        ('needs-questionnaire', 'Needs Questionnaire'),
+        ('needs-group-form', 'Needs Group Form'),
+        ('incomplete', 'Needs Submission'),
+        ('submitted', 'Submitted'),
+        ('needs-research', 'Needs Research'),
+        ('needs-staff-review', 'Needs Staff Review'),
+        ('under-review', 'Under Review'),
```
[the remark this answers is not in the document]
cjmabry: David needs it.

nominations/models.py
```diff
@@ -321,20 +447,9 @@ class Application(models.Model):
...
-        if not self.nomination:
-            self.nomination = Nomination.objects.create()
-
-        if not self.questionnaire:
-            self.questionnaire = Questionnaire.objects.create()
-
-        if self.status == 'submitted' and self.submitted_dt is None:
-            self.submitted_dt = datetime.datetime.now()
+    def auto_populate_research_fields(self):
+        """Auto-populate staff write-up fields from already present info"""
```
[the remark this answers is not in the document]
cjmabry: PEP recommendation https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

nominations/models.py
```diff
@@ -38,14 +38,18 @@ class Nomination(models.Model):
...
             return 'Nomination ' + str(self.pk)
 
     def save(self, *args, **kwargs):
-        # self.cleaned_data['status'] = 'complete'
         super(Nomination, self).save(*args, **kwargs)
+
+        # save the application to update statuses and do other conditional logic
+        # if the nomination has an application, save that application
+        if hasattr(self, 'application'):
+            self.application.save()
```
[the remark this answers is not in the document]
cjmabry: We do not use transaction blocks to my knowledge, unless they're built into django ORM somewhere.

We have three main models, which definitely can be revisited in future, but consist of:
* Nomination form (nomination)
* Questionnaire (questionnaire)
* Application (application)

An application consists of a nomination form (nomination) with various questions and responses and a questionnaire (questionnaire), consisting of various questions and responses.
 
I'd have to revisit the original implementation but I believe the thinking was logically there were three separate pieces to the puzzle and we didn't quite know what those pieces would contain exactly. Now that we're a little further along with process, I think nomination could be included in the application model, so perhaps we'll make a ticket for that.

nominations/models.py
```diff
@@ -130,10 +134,14 @@ def __unicode__(self):
 
     def save(self, *args, **kwargs):
         super(Questionnaire, self).save(*args, **kwargs)
+
         if self.response_set.count() == 0:
             for q in Question.objects.all():
                 self.response_set.create(question=q)
 
+        if self.application_set.all().exists():
+            # TODO: support multiple candidates
```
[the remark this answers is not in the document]
cjmabry: Misworded. `# TODO: explore supporting multiple applications per questionnaire`

nominations/models.py
```diff
@@ -130,10 +134,14 @@ def __unicode__(self):
 
     def save(self, *args, **kwargs):
         super(Questionnaire, self).save(*args, **kwargs)
+
         if self.response_set.count() == 0:
             for q in Question.objects.all():
                 self.response_set.create(question=q)
 
+        if self.application_set.all().exists():
```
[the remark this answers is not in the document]
cjmabry: Adding more comments.

`application_set` is how we access applications attached to the questionnaire because questionnaire is a foreign key.

nominations/models.py
```diff
@@ -355,6 +470,76 @@ def save(self, *args, **kwargs):
...
+        if not self.nomination:
+            self.nomination = Nomination.objects.create()
+
+        if not self.questionnaire:
+            self.questionnaire = Questionnaire.objects.create()
+
+    def generate_application_status(self):
+        """Returns a generated status based on completion of various items."""
+
+        if self.nomination.status == 'incomplete':
```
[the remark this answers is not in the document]
cjmabry: Will add comments on the nomination model explaining what it is. In future we can move under application but for now the reason we have two statuses is because they're separate models.

nominations/models.py
cjmabry: Front-end users, i.e. groups. Will change wording.

nominations/models.py
cjmabry: It is just inverse, there was a bug with just using inverse but I'll work through it since it is what should be done.

nominations/templates/application_status.html
cjmabry: 👍

nominations/tests.py
cjmabry: It's mocked. https://docs.djangoproject.com/en/1.8/topics/testing/overview/#the-test-database

nominations/tests.py
cjmabry: Yes, these aren't true unit tests. Am rewriting tests now.

nominations/models.py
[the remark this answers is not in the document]
ericbroder: Thanks for explaining. I think it makes sense given there are 3 parties accessing different kinds of related data - group/nomination, candidate/questionnaire, staff/application.

nominations/tests.py
ericbroder: Sounds like it is a real db that is separate from the regular app db?
```
Tests that require a database (namely, model tests) will not use your “real” (production) database. Separate, blank databases are created for the tests.

Regardless of whether the tests pass or fail, the test databases are destroyed when all the tests have been executed.

You can prevent the test databases from being destroyed by using the test --keepdb option. This will preserve the test database between runs. If the database does not exist, it will first be created. Any migrations will also be applied in order to keep it up to date.

The default test database names are created by prepending test_ to the value of each NAME in DATABASES. When using SQLite, the tests will use an in-memory database by default (i.e., the database will be created in memory, bypassing the filesystem entirely!). The TEST dictionary in DATABASES offers a number of settings to configure your test database. For example, if you want to use a different database name, specify NAME in the TEST dictionary for any given database in DATABASES.

On PostgreSQL, USER will also need read access to the built-in postgres database.

Aside from using a separate database, the test runner will otherwise use all of the same database settings you have in your settings file: ENGINE, USER, HOST, etc. The test database is created by the user specified by USER, so you’ll need to make sure that the given user account has sufficient privileges to create a new database on the system.
```

nominations/models.py
```diff
@@ -192,58 +218,180 @@ class Application(models.Model):
         choices=primary_election_type_choices,
         null=True,
     )
-    questionnaire = models.ForeignKey(Questionnaire, on_delete=models.SET_NULL, null=True, blank=True)
+    questionnaire = models.OneToOneField(
```
ericbroder: Are you sure this is a safe migration on prod db?

nominations/models.py
```diff
@@ -365,7 +512,9 @@ def save(self, *args, **kwargs):
         if self.status == 'submitted' and self.submitted_dt is None:
```
ericbroder: Isn't this a duplicate save method?

nominations/models.py
```diff
@@ -386,6 +535,64 @@ def save(self, *args, **kwargs):
...
+            status = self.status
+        return status
+
+    def is_editable(self):
+        """Returns whether a group can edit this application."""
+
+        if self.status in self.EDITABLE_STATUSES:
+            return True
+        else:
+            return False
```
ericbroder: I think this can just be `return self.status in self.EDITABLE_STATUSES`

ericbroder:
@cjmabry good work. main concern is the potential duplicate save method.

nominations/models.py
cjmabry: Aw man something messed up in merge with master, will fix.

nominations/models.py
cjmabry: From a data standpoint yes, because a OneToOneField in Django is like a ForeignKey with `unique=True` based on its implementation.

There is one other difference, and that is a foreign key returns a queryset when trying to reference it's attached objects (i.e. `application_set`) whereas a one-to-one field will return a single object. This is the one issue we may run into with A/B deploy - if code is referencing `questionnaire.application_set`, we might run into a temporary issue. This is used only here - https://github.com/Our-Revolution/site/blob/207b2515040d53a98f36d7fb7f3757d1799300e2/nominations/models.py#L124.

Am thinking through A/B deploy issues now.

ericbroder: @cjmabry if the migrations add a uniqueness constraint that wasn't there before, it might throw error if prod has duplicate data already in table?

Thanks for checking about A/B deploy issues. Also this seems out of scope for this ticket, so I'm not sure why this schema change is happening here? Might be best to back it out at this point if not needed for ticket.

cjmabry: You're right, I need sleep.

Removing!

nominations/models.py
```diff
@@ -410,6 +546,64 @@ def save(self, *args, **kwargs):
...
+
+        if self.status in self.EDITABLE_STATUSES:
+            if self.nomination.status == 'incomplete':
+                if self.questionnaire.status == 'complete':
+                    status = 'needs-group-form'
+                else:
+                    status = 'needs-group-form-and-questionnaire'
+            else:
+                # nomination complete
+                if self.questionnaire.status == 'incomplete':
```
ericbroder: @cjmabry i think there is a bug here - what about if questionnaire status is `sent`? This should be `if self.questionnaire.status != 'complete':`?

PR #129, opened 2018-02-23
TECH-825 add stand out information field

ericbroder:


ericbroder:
Ready for review but dont merge until freeze is over.

PR #130, opened 2018-02-27
TECH-826 add structured candidates field to application model

ericbroder:


nominations/templates/admin/application_pdf.html
```diff
@@ -99,12 +99,69 @@ <h2>
...
+
+      <section>
+
         <h2>
           State of the Race
         </h2>
 
         {{ app.state_of_the_race|safe }}
 
+      </section>
```
cjmabry: My understanding is that David is generating these manually right now so a feature flag here isn't required, but can we verify that is the case before we ship PDF changes for something vols won't be using yet.

nominations/views.py
```diff
@@ -507,7 +507,10 @@ class ApplicationPDFView(PDFTemplateView):
 
     def get_context_data(self, **kwargs):
         app_id = self.request.GET.get('id')
-        app = get_object_or_404(Application, pk=app_id)
+        app = get_object_or_404(
+            Application.objects.select_related(),
```
cjmabry: Nice.

nominations/templates/admin/application_pdf.html
ericbroder: I spoke to David about this, and he said only the admin UI field for volunteers needs to be gated, not pdf changes. Although volunteers have access to the pdf, my impression is that it is mainly just used by staff.

nominations/views.py
ericbroder: I find this a little confusing but glad it looks right to you.

PR #137, opened 2018-03-02
TECH-891 add confirmation for bulk status changes

ericbroder:


nominations/admin.py
```diff
@@ -1,12 +1,18 @@
+from django.conf.urls import include, url
 from collections import OrderedDict
 from django.conf import settings
 from django.contrib import admin
+from django.urls import reverse_lazy
```
cjmabry: Not used, remove.

nominations/admin.py
```diff
@@ -1,12 +1,18 @@
+from django.conf.urls import include, url
```
cjmabry: `include` not used, remove.

nominations/admin.py
```diff
@@ -1,12 +1,18 @@
...
 from django.contrib import admin
+from django.urls import reverse_lazy
 from datetime import datetime
 from django.db.models import Q
 from .models import *
+from .admin_views import ApplicationPDFView, ApplicationsStatusChangeView
 from local_groups.models import Group
 from local_groups.actions import export_as_csv_action
 import pprint
+from django.contrib.contenttypes.models import ContentType
```
cjmabry: Not used, remove.

nominations/admin.py
cjmabry: `include` not used, remove.

nominations/admin.py
cjmabry: Not used, remove.

nominations/models.py
```diff
@@ -617,6 +617,12 @@ def save(self, *args, **kwargs):
         super(Application, self).save(*args, **kwargs)
 
     class Meta:
+        permissions = (
```
cjmabry: Nice.

ericbroder: Yeah I think we should do more of this. We have lots of hardcoded access control and some of it should probably use permissions instead.

ericbroder:
@cjmabry I suggest in case like this better to approve PR but don't merge, and then I can make minor changes and merge. "Request change" function I think is better for critical issues, not minor stuff like unused packages.

PR #141, opened 2018-03-05
TECH-881 Users can't add or create a group on mobile map

cjmabry:


ourrevolution/settings.py
```diff
@@ -270,11 +270,13 @@
 """ID of 'Briefly list important local issues...' Question"""
 NOMINATIONS_QUESTION_ISSUES_ID = 24
 
+ADD_GROUP_URL = '/groups/new/'
 ORGANIZING_GUIDES_URL = '/docs/organizing-guides/'
 ORGANIZING_DOCS_URL = '/docs/'
 RESULTS_URL = '/results/'
 RESULTS_2016_URL = '/election-2016/'
 RESULTS_2017_URL = '/2017-elections-results/'
+START_GROUP_URL = 'https://docs.google.com/document/d/1BWp6HCZ6tngr6SJHJB3H1uPTX2Hcv6cMUQondCrBkHg/edit'
```
ericbroder: nice use of centralized configs 👍

pages/templates/pages/group_index_page.html
```diff
@@ -23,16 +19,13 @@ <h3 class="mt0 mb0 groups-map__heading">Local Groups <span class="glyphicon glyp
...
         <input type="search" class="form-control maps-auto-complete" placeholder="Search for a city" id="autocomplete-input">
       </div>
 
       <div class="groups-map-info__status text-left">
         <div class="component pa20 hidden-xs">
           <div class="groups-map__add">
-            <h4 class="mb20 mt0">No group in your area?</h4>
-            <a href="https://docs.google.com/document/d/1BWp6HCZ6tngr6SJHJB3H1uPTX2Hcv6cMUQondCrBkHg/edit" class="btn btn-block btn-primary uppercase ls2" target="_blank" onclick="trackOutboundLink('https://docs.google.com/document/d/1BWp6HCZ6tngr6SJHJB3H1uPTX2Hcv6cMUQondCrBkHg/edit', true)">Start a group</a>
-            <a href="/groups/new" class="btn btn-block btn-secondary uppercase ls2">Add your group</a>
+            {% include 'partials/add_group.html' %}
```
ericbroder: good use of partials 👍

pages/templates/pages/group_index_page.html
```diff
@@ -42,6 +35,10 @@ <h4 class="mb20 mt0">No group in your area?</h4>
   </div>
 </div>
 
+<div class="component pa20 visible-xs f5f5f5-bg">
```
ericbroder: @cjmabry actually one issue is I suggest we be more consistent about how we use the bootstrap grid pattern. Generally should have container/row/col pattern for layout, right? I don't understand what a component is.

ericbroder: Suggest to follow docs here http://getbootstrap.com/docs/3.3/css/#grid
```
Rows must be placed within a .container (fixed-width) or .container-fluid (full-width) for proper alignment and padding.
Use rows to create horizontal groups of columns.
Content should be placed within columns, and only columns may be immediate children of rows.
...
```

ericbroder:
May want to hold off merge until after deploy.

pages/templates/pages/group_index_page.html
```diff
@@ -42,6 +35,14 @@ <h4 class="mb20 mt0">No group in your area?</h4>
   </div>
 </div>
 
+<div class="container">
+  <div class="row">
+    <div class="col-xs-12 visible-xs f5f5f5-bg pa20">
```
ericbroder: Looks like standard grid now 😄 

One thing I'm not sure about - should `visible-xs` class be on the whole container?

ericbroder: Oh interesting, this format is deprecated: http://getbootstrap.com/docs/3.3/css/#responsive-utilities-classes

Should be `visible-xs-block` now?

PR #142, opened 2018-03-06
TECH-818 update link for submit nomination email

ericbroder:


ericbroder:
May want to hold off merge until after deploy.

PR #143, opened 2018-03-06
TECH-895 fix js error on results sort

cjmabry:


pages/templates/pages/election_tracking_page.html
```diff
@@ -132,7 +132,7 @@ <h3>No Matching Results Found</h3>
 
         {% include "partials/past_results.html" %}
 
-        <script src="//cdnjs.cloudflare.com/ajax/libs/list.js/1.2.0/list.min.js"></script>
+        <script src="//cdnjs.cloudflare.com/ajax/libs/list.js/1.5.0/list.min.js"></script>
```
ericbroder: I'm surprised there are no other changes, because I think this is a significant upgrade. For example I think fuzzy search is not a separate plugin anymore as of this version:
http://listjs.com/docs/plugins/fuzzysearch/
https://github.com/javve/list.fuzzysearch.js#deprecated-since-listjs-150

Also I briefly attempted to use latest version in my local dev when I updated this recently, and I had errors, so that's why I kept version at `1.2.0`. So unless I was mistaken there are other errors that show up with just a simple upgrade and no refactors. Did you test this thoroughly?

cjmabry: I did test everywhere list.js was used and no issues, but will look into removing fuzzy search external.

pages/templates/pages/election_tracking_page.html
```diff
@@ -131,9 +131,9 @@ <h3>No Matching Results Found</h3>
         </div>
 
         {% include "partials/past_results.html" %}
-
-        <script src="//cdnjs.cloudflare.com/ajax/libs/list.js/1.2.0/list.min.js"></script>
-        <script src="https://cdnjs.cloudflare.com/ajax/libs/list.fuzzysearch.js/0.1.0/list.fuzzysearch.js"></script>
+        
+        {# TODO: remove vendor folder from js #}
```
ericbroder: why not just do that now?

pages/static/src/js/app.js
```diff
@@ -1,5 +1,4 @@
 window.$ = window.jQuery = require('jquery');
-window.ListFuzzySearch = require('./vendor/list.fuzzysearch.min.js');
```
ericbroder: remove file too?

ericbroder:
looks good 👍

pages/static/src/js/app.js
cjmabry: Didn't want to due to caching reasons, if a user has old page and it tries to access vendor folder we'll error out in the off chance that's not cached.

ericbroder: if a user has old page then don't they also have the old js files?

PR #145, opened 2018-03-07
TECH-721 add change password page to organizing hub

ericbroder:


local_groups/views.py
```diff
@@ -163,6 +156,80 @@ def post(self, request, *args, **kwargs):
...
+        old_password = form.cleaned_data['old_password']
+        checkCredentialsResult = bsdApi.account_checkCredentials(
+            username,
+            old_password
+        )
+
+        '''
+        Should get 200 response and constituent record
+
+        https://cshift.cp.bsd.net/page/api/doc#---------------------check_credentials-----------------
```
cjmabry: Random comment - the way BSD structures its API URLs is one of the more ridiculous things I've seen.

PR #147, opened 2018-03-08
TECH-817 Allow user to resend candidate questionnaire

cjmabry:


nominations/views.py
```diff
@@ -325,22 +361,40 @@ def candidate_login(request):
 
     return render(request, 'candidate/login.html', {'form': form})
 
+
 def reset_questionnaire(request):
     app_id = request.GET.get('id')
     user = request.session['profile']
     user_id = user['user_id']
+    default_next_url = '/groups/nominations/questionnaire/edit?id=' + app_id
```
ericbroder: suggest to use named url mappings for this

nominations/views.py
```diff
@@ -325,22 +361,40 @@ def candidate_login(request):
...
+            request,
+            QUESTIONNAIRE_NOT_FOUND_ERROR
+        )
+        return redirect('/groups/nominations/dashboard/')
 
+    # get the application attached to the questionnaire
     questionnaire.status = 'incomplete'
+    application.authorized_email = None
     questionnaire.save()
+    application.save()
```
ericbroder: app save already calls questionnaire save too? don't need to do both?

nominations/templates/questionnaire_index.html
```diff
@@ -43,7 +43,10 @@ <h2 class="text-center mb20 mt20">- OR -</h2>
               <div class="col-md-12 text-center fs18 pa20 br3 f5f5f5-bg mb20">
                 <p>You've sent the candidate questionnaire to <strong>{{application.authorized_email}}</strong> - we'll notify you at <strong>{{application.rep_email}}</strong> when the candidate has completed it.</p>
                 
-                <p>If your candidate is having trouble with the nominations platform, <a href="/groups/nominations/questionnaire/reset?id={{application.id}}">click here</a> to fill out the questionnaire on behalf of your candidate.</p>
+                <p>If your candidate is having trouble with the nominations platform, you can resend the questionnaire or fill it out yourself on behalf of your candidate using the buttons below.</p>
+                
+                <a href="/groups/nominations/questionnaire/reset?id={{application.id}}&next=/groups/nominations/questionnaire?id={{application.id}}" class="btn btn-primary btn-block uppercase ls2">Resend questionnaire</a>
+                <a href="/groups/nominations/questionnaire/reset?id={{application.id}}" class="btn btn-secondary btn-block uppercase ls2">Fill out manually</a>
```
ericbroder: Suggest to use named url mappings for links.

ericbroder:
Main concern is the extra save call. Otherwise looks good 👍

nominations/views.py
cjmabry: This is needed because while questionnaire saves the application in the questionnaire save method, authorized_email changing would not be saved in that method. Thus, an applicaiton save here is needed to save authorized_email change.

ericbroder: I suggest to use a solution where application is saved only once. For example could do something like this where `skip_application` is a param that will skip the auto app save:
```
    questionnaire.save(skip_application=True)
    application.authorized_email = None
    application.save()
```

cjmabry: Like this, will implement.

cjmabry: Done.

nominations/models.py
```diff
@@ -150,7 +150,7 @@ def _campaign_issues(self, *args, **kwargs):
         return position
     campaign_issues = property(_campaign_issues)
 
-    def save(self, *args, **kwargs):
+    def save(self, skip_application_save=False, *args, **kwargs):
```
ericbroder: I wonder if it would be better to use the existing `**kwargs` param to pass in `skip_application_save`? This way you don't have to change the method signature at all. Although I am still learning how to use `*args` and `**kwargs` and the concept of undefined and countless potential params seems kind of bizarre. What is the point of having a method signature contract if it's also like "by the way feel free to pass in any number of other params too"? In the way you have it now at least it is obvious at a glance that there is a `skip_application_save` param.

nominations/views.py
```diff
@@ -325,22 +361,41 @@ def candidate_login(request):
...
+        messages.error(
+            request,
+            QUESTIONNAIRE_NOT_FOUND_ERROR
+        )
+        return redirect('/groups/nominations/dashboard/')
 
     questionnaire.status = 'incomplete'
-    questionnaire.save()
+    application.authorized_email = None
+    questionnaire.save(skip_application_save=True)
```
ericbroder: nice 👍

nominations/models.py
cjmabry: Yeah, this is a bit confusing, and I ultimately went this route to illustrate to dev that this was indeed a thing.

PR #149, opened 2018-03-08
TECH-905 update OR address

ericbroder:


ericbroder:
@cjmabry what is the point of changing those? It functions fine as placeholder examples so that's why I left alone. This is not technically us displaying our mailing address, it's just using an address which happens to be our old office to just display example text for address related form fields.

ericbroder:
Updated from feedback.

PR #151, opened 2018-03-12
TECH-657 automatically clear cache when endorsement is updated

ericbroder:


pages/models.py
```diff
@@ -986,6 +1019,11 @@ def news_published_handler(instance, **kwargs):
     news_post_changed(instance)
 
 
+@receiver(page_unpublished, sender=NewsPost)
+def news_unpublished_handler(instance, **kwargs):
```
cjmabry: Good call.

PR #161, opened 2018-03-21
TECH-821 CC group leader when the candidate questionnaire is sent

cjmabry:


nominations/views.py
```diff
@@ -233,17 +237,31 @@ def form_valid(self, form):
...
 
-        subject="You're being nominated for endorsement by an official Our Revolution group!"
-        from_email='Our Revolution <info@ourrevolution.com>'
-        to_email=["%s <%s>" % (candidate_name,candidate_email)]
+        subject = "You're being nominated for endorsement by an official Our Revolution group!"
+        from_email = 'Our Revolution <info@ourrevolution.com>'
+        to_email = ["%s <%s>" % (candidate_name, candidate_email)]
+        cc_emails = [
+            "%s <%s>" % (group, rep_email),
+            "%s <%s>" % (group, settings.ELECTORAL_COORDINATOR_EMAIL)
```
ericbroder: Why is `group` used here? Typo?

ourrevolution/settings.py
```diff
@@ -279,6 +279,10 @@
 NOMINATIONS_QUESTION_ISSUES_ID = 24
 
 ADD_GROUP_URL = '/groups/new/'
+ELECTORAL_COORDINATOR_EMAIL = os.environ.get(
+    'ELECTORAL_COORDINATOR_EMAIL',
+    'political@ourrevolution.com'
```
ericbroder: It's good we are not putting anyone's personal email into the codebase :)

I suggest to make a setting for political@ourrevolution.com if we don't have one already.

nominations/views.py
cjmabry: Oops good catch.

cjmabry:
@ericbroder fixed.

ourrevolution/settings.py
```diff
@@ -279,6 +279,10 @@
 NOMINATIONS_QUESTION_ISSUES_ID = 24
 
 ADD_GROUP_URL = '/groups/new/'
+ELECTORAL_COORDINATOR_EMAIL = str(os.environ.get(
```
ericbroder: why do we need `str` here? we don't use it elsewhere.

cjmabry: To allow `.env` to be 'email@example.com' or email@example.com and still work.

nominations/views.py
```diff
@@ -233,17 +237,34 @@ def form_valid(self, form):
...
 
         text_content = plaintext.render(d)
         html_content = htmly.render(d)
-        msg = EmailMultiAlternatives(subject, text_content, from_email, to_email)
+        msg = EmailMultiAlternatives(
+            subject,
+            text_content,
+            from_email,
+            to_email,
+            cc=cc_emails
```
ericbroder: @cjmabry not sure why this is not working? Seems like should probably support list value? https://github.com/django/django/blob/stable/1.10.x/django/core/mail/message.py

PR #166, opened 2018-03-23
TECH-969 add organizing hub dashboard page to cms

ericbroder:


local_groups/templatetags/local_groups_tags.py
```diff
@@ -10,11 +10,6 @@ def bsd_create_account_url():
     return settings.BSD_CREATE_ACCOUNT_URL
 
 
-@register.simple_tag
```
cjmabry: This is being removed because we no longer use it, instead linking to our internal page, correct?

PR #178, opened 2018-04-03
TECH-805 remove old config for organizing hub

ericbroder:


local_groups/urls.py
```diff
@@ -16,74 +16,74 @@
...
-                        name='password_reset_done',
-                    ),
-                ])),
-            ])),
-            url(r'^verify-email/', include([
                 url(
-                    r"^confirm/(?P<key>[-:\w]+)/$",
-                    VerifyEmailConfirmView.as_view(),
-                    name="groups-verify-email-confirm"
+                    r'^confirm/(?P<uidb64>[0-9A-Za-z_\-]+)/(?P<token>[0-9A-Za-z]{1,13}-[0-9A-Za-z]{1,20})/$',
```
cjmabry: What's happening here? Wasn't this just a config change?

ericbroder: This should all just be an indentation change. Github is no bueno at displaying trivial indent changes 😞

local_groups/urls.py
```diff
@@ -16,74 +16,74 @@
     url(r'^join-us-on-slack', SlackInviteView.as_view())
 ]
 
-if settings.BSD_LOGIN_ENABLED:
```
ericbroder: this is the main change, removing this line, then un-indenting the stuff below

local_groups/urls.py
cjmabry: Oh Okay I see now, thanks for clarifying!

PR #219, opened 2018-05-10
TECH-1125 add staff recommendation field

ericbroder:


nominations/models.py
```diff
@@ -468,6 +474,7 @@ class Application(models.Model):
         null=True,
     )
 
+    """TODO: remove?"""
```
cjmabry: @ericbroder Can we create followup ticket to remove.

ericbroder: https://ourrevolution.atlassian.net/browse/TECH-1138

PR #223, opened 2018-05-16
TECH-1142 add primary wins to results page

ericbroder:


ericbroder:
dont merge to master until there is further review/signoff

pages/static/src/scss/components/_card.scss
```diff
@@ -67,6 +67,14 @@
       margin-left: 10px;
     }
 
+    &.faded {
```
cjmabry: Wondering if this is necessary. Not quite sure. Why do we need to distinguish outside of Won Primary?

ericbroder: Well overall of course it just adds to making a distinction so it's clear there is a different. But also the status here as a "won" is temporary for primaries, because if they lose in the general this result will change to a loss. whereas the others are permanent results.

of course this also raises the question of how we indicate primary wins when the candidate loses general, but for now i am assuming that we do not.

cjmabry: Thinking about it more I'm fine with opacity change.

As for second question I think we only show the latest results, i.e. primary win, loss, or general election win, and I think this is how that will function, correct?

ericbroder: yes, basically with this change there are 3 mutually exclusive result categories a candidate can be in: won (general), won primary (primary), loss (general or primary)

PR #227, opened 2018-05-18
TECH-738 upgrade wagtail

ericbroder:


requirements.txt
```diff
@@ -65,7 +65,7 @@ unicodecsv==0.14.1
 virtualenv==15.0.3
 virtualenv-clone==0.2.6
 virtualenvwrapper==4.7.2
-wagtail==1.7
+wagtail>=1.8,<1.9
```
cjmabry: What's the reason we don't just use `==1.8` - is it because we want the latest minor version?

ericbroder: We want the latest patch version. The minor version is `1.8.x`, which won't change. But we want patch/bug fixes that are applied to the minor version we are on.

ericbroder: see: http://docs.wagtail.io/en/v1.13.1/releases/upgrading.html#upgrade-process

ericbroder:
dont merge until code freeze is over

requirements.txt
cjmabry: Cool, that's what I assumed, thanks!

cjmabry:
Don't merge until after today's deploy.

PR #228, opened 2018-05-21
TECH-1166 Temporarily remove list of staff names from website

cjmabry:


pages/templates/pages/about_page.html
```diff
@@ -96,29 +96,33 @@ <h1>Membership</h1>
 	<section class="row pt4 pb4" id="board">
 		<div class="container">
 			<div class="row">
-				<div class="col-md-6 pb2">
+				<div class="col-md-6 {{ page.board_list|yesno:",col-md-offset-3" }} pb2">
```
ericbroder: why adding a comma here? is that a typo?

pages/templates/pages/about_page.html
```diff
@@ -96,29 +96,33 @@ <h1>Membership</h1>
...
+				{% endif %}
 			</div>
 		</div>
 	</section>
 
 	<section class="blue-bg white-text row pt4 pb4" id="staff">
 		<div class="container">
 			<div class="row">
-				<div class="col-md-6 pb2">
+				<div class="col-md-6 {{ page.staff_list|yesno:",col-md-offset-3" }} pb2">
```
ericbroder: ditto

pages/templates/pages/about_page.html
ericbroder: Ok I see now: https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#yesno

Although looks like `yesno` is usually used for boolean values. In this case I would expect an if statement, like `{% if not page.board_list %}`. However if you tested and it works, this is fine too.

PR #231, opened 2018-05-25
TECH-1170 add model for local group affiliations

ericbroder:


local_groups/models.py
```diff
@@ -455,3 +457,32 @@ def save(self, *args, **kwargs):
...
         return self.name
+
+
+class LocalGroupProfile(models.Model):
+    """Local group information for a user"""
+
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
+
+    def __unicode__(self):
+        return str(self.user.id) + ": " + self.user.email
```
cjmabry: This might throw errors since email isn't required by default on a user, although I assume we'll enforce it elsewhere for this model.

ericbroder: @cjmabry this doesn't throw an error when email is blank, it just shows up blank here.

All bsd users should have an email address because they log in with email and then we attach it to their user record. So the only users I see without email addresses are non-BSD users - both staff and volunteers. And for these cases, (1) they shouldn't have local group affiliation or be using the organizing hub, and (2) I think they should probably have email addresses on those user account?

I don't see the purpose of not putting an email address on staff or volunteer user accounts. This means the user is harder to identify and they can't use password reset or get email notifications?

cjmabry: @ericbroder If it doesn't throw an error then completely fine.

We don't have email addresses on some staff/volunteer user accounts because it will not allow them to login if they also have a BSD account. I believe we have a ticket for this to fix. But further, it's just not required on django's end.

ericbroder: @cjmabry ok thanks. I don't see ticket for the other issue you mention, please add to jira if not there already.

PR #235, opened 2018-05-31
TECH-1169 fix broken meta image

ericbroder:


pages/templates/pages/base_page.html
```diff
@@ -18,13 +18,9 @@
             <meta property="og:description" content="{% firstof page.search_description "The next step for Bernie Sanders' movement is a new group called Our Revolution, which will fight to transform America and advance the progressive agenda that we believe in." %}" />
             <meta property="og:type" content="website" />
             <meta property="og:site_name" content="Our Revolution" />
-            {% if request.GET.category %}
-                {% with "img/dist/meta/"|add:request.GET.category|add:".png" as social_image %}
```
ericbroder: Seems like `img/dist/meta/` should be `dist/img/meta/` instead. I guess it never worked? I suspect this used to throw a 404 error, and when we enabled the versioned static assets it became a server error instead.

Also other weird issues here:
- meta image needs absolute url
- twitter meta image should match
- this special logic should be on specific pages, not the global page template

ericbroder: also there is no `all` meta image

nominations/templates/admin/application_pdf.html
```diff
@@ -2,7 +2,7 @@
 
 {% block content %}
     {% load static humanize %}
-    <link rel="stylesheet" href="{% static "dist/css/main.min.css" %}?v=122">
```
ericbroder: @cjmabry is there any reason this change would case an error?

ericbroder: See: https://github.com/Our-Revolution/site/pull/237

PR #241, opened 2018-06-05
TECH-1199 Add endorsement process to the site

cjmabry:


pages/templatetags/pages_tags.py
```diff
@@ -20,6 +20,11 @@ def candidates_url():
     return settings.CANDIDATES_URL
 
 
+@register.simple_tag
+def endorsement_process_url():
```
ericbroder: I wonder if this makes more sense in the `nominations` app/tags?

cjmabry: Hmmm. I'm not sure actually, I could see it going either way. Since past code freeze going to merge but I think either is fine.

PR #243, opened 2018-06-06
TECH-1174 refactor old access control in hub to start using new access control

cjmabry:


local_groups/forms.py
```diff
@@ -152,8 +152,8 @@ class GroupLoginForm(AuthenticationForm):
         widget=forms.TextInput(attrs={'autofocus': True}),
         help_text='''
             This should be the official group leader email address registered
-            with Our Revolution, or the email you use for submitting
-            nominations
+            with Our Revolution, a group admin email, or the email you use for
```
ericbroder: Can we also change label to something more generic than `Group Admin Email`?

local_groups/views.py
```diff
@@ -40,7 +41,12 @@
 
 
 @method_decorator(verified_email_required, name='dispatch')
```
ericbroder: Should remove the decorator, it's redundant.

local_groups/views.py
```diff
@@ -49,18 +55,32 @@ class EventCreateView(LoginRequiredMixin, SuccessMessageMixin, CreateView):
...
     Your event was created successfully and is now being reviewed by our team.
     '''
+    permission_required = 'local_groups.add_event'
+
+    def get_local_group(self):
+        local_group = None
+        user = self.request.user
 
-    # Check if user is a group leader and has a valid bsd cons_id
+        if hasattr(user, 'localgroupprofile'):
```
ericbroder: I liked your suggestion of putting this into a reusable method. Although I did not implement either because was out of time.

local_groups/views.py
```diff
@@ -49,18 +55,32 @@ class EventCreateView(LoginRequiredMixin, SuccessMessageMixin, CreateView):
...
-        is_group_leader = Group.objects.filter(
-            rep_email__iexact=self.request.user.email,
-            status__exact='approved',
-        ).first() is not None
         user = self.request.user
         has_valid_cons_id = hasattr(user, 'bsdprofile') and (
             user.bsdprofile.cons_id != BSDProfile.cons_id_default
         )
-        return is_group_leader and has_valid_cons_id
+        return has_valid_cons_id
```
ericbroder: Oh right I forgot about this part, guess you can leave this method in for now.

local_groups/views.py
```diff
@@ -121,44 +141,25 @@ def post(self, request, *args, **kwargs):
 
 # View for Admin updates to Group Info
 @method_decorator(verified_email_required, name='dispatch')
```
ericbroder: rm

organizing_hub/mixins.py
```diff
@@ -45,7 +45,7 @@ def has_permission(self):
         permissions = self.get_permission_required()
 
         """Return False for non-approved local group"""
-        if local_group.status != 'approved':
+        if local_group and local_group.status != 'approved':
```
ericbroder: Why change? I think `local_group` can never be `None`, otherwise would have thrown exception in get method?

cjmabry: When we override the get_local_gorup method if a group is set to none it makes it to here and throws an error. I was unsure how to fix without adding this.

local_groups/forms.py
cjmabry: 👍

local_groups/views.py
cjmabry: Same, we can if needed in future, I didn't want to do because wasn't sure what you were doing.

local_groups/views.py
cjmabry: Yeah I was unsure how to handle this so just left it in.

local_groups/forms.py
ericbroder: @cjmabry thanks! I noticed this page also has the title/page title with `Group Admin Login`. Can we change that to be generic too?

PR #248, opened 2018-06-13
TECH-739 upgrade wagtail to 1.9

cjmabry:


requirements.txt
```diff
@@ -26,12 +26,12 @@ django-debug-toolbar==1.6
 django-dotenv==1.4.1
 django-easy-pdf==0.1.1
 django-localflavor==1.4.1
-django-modelcluster==2.0
+django-modelcluster==3.0
```
ericbroder: can we use the syntax that will upgrade to latest minor version or bug fix version or something like that? i don't see why we should pin it to the x.0 version. for wagtail we are pulling in the latest minor version for a specific major version.

see related discussion: https://github.com/Our-Revolution/site/pull/227/files#r189662101

requirements.txt
```diff
@@ -26,12 +26,12 @@ django-debug-toolbar==1.6
...
 django-multiselectfield==0.1.5
 django-phonenumber-field==1.1.0
 django-recurrence==1.4.1
 django-storages==1.5.1
-django-taggit==0.18.3
+django-taggit==0.20
```
ericbroder: same here, would like latest minor/patch version?

requirements.txt
ericbroder: for example i see a 3.0.1 version with a bug fix available

requirements.txt
ericbroder: lots of more recent versions available: ```
0.22.2 (2017-12-27)
0.22.1 (2017-04-22)
0.22.0 (2017-01-29)
0.21.6 (2017-01-25)
0.21.5 (2017-01-21)
0.21.4 (2017-01-10)
0.21.3 (2016-10-07)
0.21.2 (2016-08-31)
0.21.1 (2016-08-25)
0.21.0 (2016-08-22)
0.20.2 (2016-07-11)
0.20.1 (2016-06-23)```

requirements.txt
```diff
@@ -26,12 +26,12 @@ django-debug-toolbar==1.6
...
 django-easy-pdf==0.1.1
 django-localflavor==1.4.1
-django-modelcluster==2.0
+django-modelcluster>=3.0,<=4.0
 django-multiselectfield==0.1.5
 django-phonenumber-field==1.1.0
 django-recurrence==1.4.1
 django-storages==1.5.1
-django-taggit==0.18.3
+django-taggit>=0.20,<=0.21
```
ericbroder: should be `<` not `<=`?

requirements.txt
```diff
@@ -26,12 +26,12 @@ django-debug-toolbar==1.6
 django-dotenv==1.4.1
 django-easy-pdf==0.1.1
 django-localflavor==1.4.1
-django-modelcluster==2.0
+django-modelcluster>=3.0,<=4.0
```
ericbroder: should be `<` not `<=`?

ericbroder: also i don't know if this is intentional or not, but the other cases are updating to a certain minor version branch such as `>= 1.0,<1.1`. however this is updating to a certain major version branch, `>=3.x,<4.x`. any reason for this?

PR #249, opened 2018-06-14
Tech 1089

dplynch:
# Remove orphan page
# https://ourrevolution.com/electoral-coordinator/


cjmabry:
Hey there @dplynch!

So if I understand this pull request correctly, you want to get rid of https://ourrevolution.com/electoral-coordinator/? If so, that page is managed in our CMS, so a code change is not required. I will take it down and redirect it, so thanks for flagging!

A few tips for if you submit more pull requests:
- Make sure you're updated to the latest master commit before submitting a pull request, this way we can see just the changes you made.
- It's probably better practice to open an issue before committing a pull request, so we can make sure it is a needed change and discuss requirements.

We're going to be working on contribution guidelines soon, so thanks for contributing!

PR #251, opened 2018-06-19
TECH-1180 add Update Event pages to hub

ericbroder:


bsd/models.py
```diff
@@ -1,10 +1,316 @@
...
+
+class BSDEvent(models.Model):
+    """BSD Event"""
+    objects = BSDEventManager()
+
+    duration_type_choices = (
+        (1, 'Minutes'),
+        (2, 'Hours'),
+    )
+    event_type_choices = (
```
cjmabry: Not required for this ticket, but is there anyway we could make this not be hardcoded?

ericbroder: we discussed. doesnt have to be hardcoded but this seems like preferred approach for now.

PR #256, opened 2018-06-22
TECH-1238 fix issues with bsd api

ericbroder:


bsd/models.py
```diff
@@ -52,8 +55,13 @@ def from_json(self, data):
             )
             venue_state_or_territory = data["venue_state_cd"]
 
-        """Fix duration format"""
-        duration_count = int(duration) if duration != '' else 0
+        """Set duration to a positive integer if possible"""
+        """TODO: support all day events (-1)?"""
```
cjmabry: If an all-day event already exists in BSD will that break anything here?

bsd/forms.py
```diff
@@ -20,7 +20,8 @@ class BSDEventForm(forms.ModelForm):
         label="Capacity Limit",
         min_value=0,
     )
-    duration_count = forms.IntegerField(min_value=0)
+    """Duration form field should be required and positive"""
```
cjmabry: Didn't we determine BSD didn't require duration?

bsd/models.py
ericbroder: If all day event exists in BSD, app will set duration to none here, and then Update form will require a new value for duration field. If we want to support all day events we need to do some more work here. Apparently hydra did not support it.

bsd/forms.py
ericbroder: Upon further testings, the api call fails if duration field is missing, and the admin UI seems to actually set a default value of duration field is missing. Also setting this value to 0 has weird result where start time doesn't display for event, so I removed that option too.

bsd/models.py
cjmabry: Okay that is fine, thanks for explaining.

bsd/forms.py
cjmabry: Gotcha. Classic BSD, thanks for explaining.

PR #257, opened 2018-06-23
TECH-1184 add admin pages for event promotion

ericbroder:


organizing_hub/views.py
```diff
@@ -372,35 +399,12 @@ def form_valid(self, form):
...
-"""
-        ).render(Context({
-            'bsd_base_url': BSD_BASE_URL,
-            'event_url': event.absolute_url,
-            'user_message': user_message,
-        }))
+        form.instance.message = event_promote_admin_message_template.render(
+            Context({
+                'event_url': event.absolute_url,
+            })) + user_message + event_promote_footer_template.render(Context({
```
ericbroder: Trying to fix a weird encoding issue with apostrophe in user message.

organizing_hub/views.py
```diff
@@ -32,9 +32,36 @@
 bsd_api = BSD().api
 
 BSD_BASE_URL = settings.BSD_BASE_URL
+
+event_promote_admin_message_template = Template("""Hi --
```
cjmabry: What's reason for these being lowercase? Is it because we're using `Template` here?

ericbroder: Just the lack of a clear standard? I think we just need some clear standards here. Usually I use lowercase by default for most things.

PR #258, opened 2018-06-25
TECH-743 fix immutability upgrade issue

ericbroder:


local_groups/forms.py
```diff
@@ -55,7 +55,7 @@ class Meta(object):
 
     def __init__(self, *args, **kwargs):
         if args:    # If args exist
-            data = args[0]
+            data = args[0].copy()  # Copy to make it mutable
```
ericbroder: @cjmabry do you understand what all this logic is for? i do not.

ericbroder: i think this fix works, unless the intended change is not being saved/persisted due to the new pattern. but im confused how to test because i dont understand what this all is supposed to do.

cjmabry: This was a custom implementation to take lat and long from form and convert it to a point that is needed to display on the map if I recall. It looks to be poorly implemented and might not even be needed. #techdebt

PR #259, opened 2018-06-26
TECH-1228 add permissions to override read only fields

ericbroder:


events/migrations/0003_auto_20180626_1841.py
```diff
@@ -0,0 +1,19 @@
...
+# Generated by Django 1.11.13 on 2018-06-26 18:41
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('events', '0002_auto_20180622_1817'),
```
cjmabry: @ericbroder Was this meant to be included on this OR? Fine if so we don't have to rework just want to make sure.

events/migrations/0003_auto_20180626_1841.py
```diff
@@ -0,0 +1,19 @@
...
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='eventpromotion',
+            options={'ordering': ['-date_submitted']},
```
ericbroder: Add missing migration from https://github.com/Our-Revolution/site/pull/257/files#diff-01923c357366c8062d06c4a3affea768R55

PR #272, opened 2018-07-11
TECH-1254 debug strict mode

ericbroder:


ericbroder:
testing on QA

scripts/install_packages.sh
```diff
@@ -1,13 +1,13 @@
...
-# set -eo pipefail
-# IFS=$'\n\t'
+set -eo pipefail
+IFS=$'\n\t'
 
 echo "Start installing packages"
 cd /home/ubuntu/ourrevolution
 source /home/ubuntu/.virtualenvs/ourrevolution/bin/virtualenvwrapper.sh
-workon ourrevolution
+workon ourrevolution || true
```
cjmabry: Why do we need `|| true` here? If environment isn't available, isn't something wrong?

ericbroder: https://github.com/Our-Revolution/site/pull/275/files

ericbroder: https://github.com/Our-Revolution/site/pull/275

PR #286, opened 2018-07-25
TECH-1307 create contact model

ericbroder:


contacts/migrations/0001_initial.py
```diff
@@ -0,0 +1,30 @@
...
+    operations = [
+        migrations.CreateModel(
+            name='Contact',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('email_address', models.EmailField(blank=True, max_length=255, null=True)),
+                ('external_id', models.CharField(blank=True, max_length=128, null=True, unique=True)),
+                ('first_name', models.CharField(blank=True, max_length=255, null=True)),
+                ('last_name', models.CharField(blank=True, max_length=255, null=True)),
+                ('point', django.contrib.gis.db.models.fields.PointField(blank=True, null=True, srid=4326)),
```
cjmabry: Do we see any value to storing zipcode? It's the most basic location information we store on pretty much all constituents and we sometimes use it to target. Although perhaps point covers us here?

ericbroder: Rather than add all the fields that might be useful in the future, I focused on adding the fields that I know we need now. So I suggest we can just iterate in the future assuming we like this pattern in general.

cjmabry: Cool, agreed.

PR #287, opened 2018-07-26
TECH-693 add create account page

ericbroder:


bsd/models.py
```diff
@@ -5,30 +5,78 @@
...
+class Account(models.Model):
+    """BSD User Account"""
+    email_max_length = 128
+    name_max_length = 128
+    password_max_length = 100
+    postal_code_max_length = 16
+    email_address = models.EmailField(max_length=email_max_length)
+    first_name = models.CharField(max_length=name_max_length)
+    last_name = models.CharField(max_length=name_max_length)
+    password = models.CharField(max_length=password_max_length)
```
cjmabry: I assume this will be storing a hashed password? Should we specify here?

ericbroder: I'm not sure what you mean, we don't store it, we are just sending it to bsd api?

cjmabry: Aren't these models stored in DB?

ericbroder: No. `managed = False` means do not store in db.

cjmabry: Ahhhh gotcha, missed that, thanks!

PR #290, opened 2018-08-03
TECH-1332 finish list generation

ericbroder:


organizing_hub/tasks.py
```diff
@@ -120,60 +75,73 @@ def sync_contact_list_with_bsd_constituents(
...
+                            'first_name': constituent.find('firstname').text,
+                            'last_name': constituent.find('lastname').text,
+                            'point': constituent_point,
+                        },
+                    )
+                    contact_list.contacts.add(contact)
+
+    """Get list limit from max contacts and point."""
+    if point is None or max_contacts is None or max_contacts == 0:
+        list_limit = None
```
cjmabry: We should never have no list_limit. Should enter a default if not empty so we avoid sending accidental massive sends.

ourrevolution/settings.py
```diff
@@ -51,6 +64,14 @@
     'EVENTS_PROMOTE_MAX',
     4000
 ))
+EVENTS_PROMOTE_MAX_DISTANCE_MILES = int(os.environ.get(
+    'EVENTS_PROMOTE_MAX_DISTANCE_MILES',
+    100
```
cjmabry: We should probably set this config to something much smaller, but this default is fine.

organizing_hub/tasks.py
ericbroder: will change

ourrevolution/settings.py
ericbroder: seems ok to me? its going to sort by distance anyway if theres too many.

cjmabry: I don't see many scenarios where people will be driving 100 miles to an event, and sending them an email for an event they'd never attend only risks an unsubscribe. We can talk about an appropriate number.

PR #317, opened 2018-08-28
TECH-1367 use smaller images files for thumbnails

cjmabry:


pages/templates/pages/index_page.html
```diff
@@ -12,23 +12,23 @@
...
+        {% image page.block_1_background_image min-1200x1200 as block_1_background_image %}
         background-image: url('{{ block_1_background_image.url }}');
       {% endif %}
     }
 
     #home-block-2 {
       background-color: #{{ page.block_2_background_color }};
       {% if page.block_2_background_image %}
-        {% image page.block_2_background_image original as block_2_background_image %}
+        {% image page.block_2_background_image min-1200x1200 as block_2_background_image %}
```
ericbroder: How did you choose these numbers to use? What is the pattern for the 1/2/3 differences? Maybe add code comments to explain?

cjmabry: The maximum width these images can all be at is 600 (1200 for retina support) because on mobile that's what their max-width is.

The maximum height they can all be is:
first block can be full height, or 600 (1200 for retina)
second block can be full height, or 600 (1200 for retina)
third block can only be 240 height (or 480 on retina)

Will make code comments.

PR #323, opened 2018-09-13
TECH-1321 send email promotion via bsd

ericbroder:


organizing_hub/signals.py
```diff
@@ -91,6 +94,31 @@ def sync_group_leader_affiliation_for_user(user):
...
+    Start task to send approved event promotion if contact list is complete and
+    list is not empty
+    """
+
+    contact_list = instance
+    if contact_list.status == ContactListStatus.complete.value[0] and (
+        contact_list.contacts.count() > 0
+    ):
+
+        """Check if contact list is attached to approved event promotion"""
```
cjmabry: Do we check anywhere that an event has been approved?

organizing_hub/signals.py
```diff
@@ -91,6 +94,31 @@ def sync_group_leader_affiliation_for_user(user):
...
+        if hasattr(contact_list, 'eventpromotion'):
+            event_promotion = contact_list.eventpromotion
+
+            """TODO: debug why this condition is not working as expected"""
```
cjmabry: Can you expand here? What's not working as expected?

organizing_hub/signals.py
ericbroder: Separate ticket

organizing_hub/signals.py
ericbroder: Sorry for confusion. The if statement appears to still let through use cases that shouldn't go through. It's a minor issue because the next method verifies the same thing again, so ultimately it's not impacting results. But I don't quite understand why this if statement doesn't work right, seems to be checking stale data perhaps. Might need to do a fresh lookup of the data from db to get the correct info?

organizing_hub/signals.py
cjmabry: 👍

organizing_hub/signals.py
cjmabry: Interesting. Okay thanks for explanation!

PR #327, opened 2018-09-18
TECH-1311 deploy celery changes gracefully

ericbroder:


ericbroder:
self merge so I can test on qa

PR #332, opened 2018-09-18
 TECH-1400 implement bsd sendableconsgroup / chapter / branch

cjmabry:


organizing_hub/tasks.py
```diff
@@ -115,7 +117,19 @@ def sync_contact_list_with_bsd_constituent(
     is_subscribed = False
     email_address = cons_email.findtext('email')
     is_subscribed = cons_email.findtext('is_subscribed') == '1'
-    if email_address is None or not is_subscribed:
+
+    """Check if member of sendable cons group if enabled"""
+    if EVENTS_PROMOTE_SENDABLE_CONS_GROUP_ENABLED:
```
ericbroder: I don't think this logic really belongs here? There is a filter for cons group: https://cshift.cp.bsd.net/page/api/doc#-----------------Filter-Specification-------------

I suggest we just filter by cons group on the initial query to BSD api, right? I don't see the point of doing it here so late in the process when we could do it up front.

ourrevolution/settings.py
```diff
@@ -86,6 +86,16 @@
...
 ))
+EVENTS_PROMOTE_SENDABLE_CONS_GROUP_ENABLED = bool(int(os.environ.get(
+    'EVENTS_PROMOTE_SENDABLE_CONS_GROUP_ENABLED',
+    0
+)))
+"""Default to -1 so we make sure we don't pull a real cons group if
+config isn't set, since cons groups start at 1"""
+EVENTS_PROMOTE_SENDABLE_CONS_GROUP_ID = int(os.environ.get(
+    'EVENTS_PROMOTE_SENDABLE_CONS_GROUP_ID',
+    -1
```
ericbroder: Can we just default to `None` here instead? That seems better than using -1 which is a bit confusing. Will `None` throw an error when casting to int?

Also why do we need a separate config for enabled or not, maybe we can just use this one config, and assume it's enabled if there's a valid value here, and disabled if there is no valid value here?

organizing_hub/tasks.py
cjmabry: @ericbroder This is if we decide to not use a sendable cons group in the future, i.e. if it is no longer maintained or used by our email program, perhaps because we cull our list and just don't keep non-sendable addresses. So we'd just disable here and be able to send form our full list. If this isn't here we can only send to members of a particular cons group.

ourrevolution/settings.py
cjmabry: You can't default to NONE when using an int field, so could remove int check here but that might not be good idea. Thoughts?

I see your point about separate enabled field, perhaps we can assume it's disable if this is missing.

organizing_hub/tasks.py
cjmabry: Oh wait I misread your comment, one sec.

cjmabry: Damn I didn't even see that filter, you're right, will change.

ericbroder: @cjmabry I don't understand, you can make it an optional param on the api call and only call it when enabled. This doesn't seem to address my point that this is in the wrong location? Why would we query all constituents first, store them in db, and then check if they are sendable later before we send, when we could just (optionally) query it all up front and not have to handle those non-sendable constituents at all?

ourrevolution/settings.py
ericbroder: Yeah not quite sure best way to do this if `int(None)` throws error. Maybe like this? ```
EVENTS_PROMOTE_SENDABLE_CONS_GROUP_ID = if os.environ.get(
    'EVENTS_PROMOTE_SENDABLE_CONS_GROUP_ID',
    None
)
if EVENTS_PROMOTE_SENDABLE_CONS_GROUP_ID is not None:
    EVENTS_PROMOTE_SENDABLE_CONS_GROUP_ID = int(
        EVENTS_PROMOTE_SENDABLE_CONS_GROUP_ID
    )
```

Comment (ericbroder) [reply ourrevolution/settings.py]: Another option is just leave it as a string? Are we sure it needs to be an int?

File: organizing_hub/tasks.py (hunk 1 above)
Comment (cjmabry) [reply organizing_hub/tasks.py]: @ericbroder Apparently BSD Python API doesn't support cons_group filter https://github.com/bluestatedigital/bsd-api-python/blob/master/bsdapi/Filters.py

File: ourrevolution/settings.py [hunk 3]
```diff
@@ -86,6 +86,10 @@
     'EVENTS_PROMOTE_RECENT_CUTOFF_DAYS',
     14
 ))
+EVENTS_PROMOTE_SENDABLE_CONS_GROUP_ID = str(os.environ.get(
```
ericbroder: `str` is not needed

ericbroder: And actually it might break None handling so please remove?

bsd/models.py
```diff
@@ -64,7 +65,12 @@ def find_constituents_by_state_cd(state_cd):
     filter = {}
     filter['state_cd'] = str(state_cd)
     filter['is_subscribed'] = True
-    bundles = ['primary_cons_addr', 'primary_cons_email']
+
+    """Check if member of sendable cons group"""
+    if EVENTS_PROMOTE_SENDABLE_CONS_GROUP_ID:
+        filter['cons_group'] = EVENTS_PROMOTE_SENDABLE_CONS_GROUP_ID
+
+    bundles = ['primary_cons_addr', 'primary_cons_email','cons_group']
```
ericbroder: why do we need cons group bundle? remove?

ourrevolution/settings.py
ericbroder: this var is named after events app but only used in bsd app. maybe update naming convention to be consistent? `BSD_`?

ericbroder: actually it makes sense for this to be a event promo settings, but it should be invoked in events app or organizing hub, and then passed in bsd app as a param to the find method, right?

bsd/models.py
```diff
@@ -45,7 +45,8 @@ def assert_valid_account(api_result):
 
 def find_constituents_by_state_cd(state_cd):
```
ericbroder: i suggest to add optional cons group param to method signature. this feels too much hardcoded logic. when we get list for calling tool, i dont think we care about sendable cons group right?

cjmabry:
@ericbroder chapter / branch stuff is not needed, outlined in research doc in ticket.

ourrevolution/settings.py
cjmabry: Done.

bsd/models.py
cjmabry: Done.

ourrevolution/settings.py
cjmabry: Agreed.

bsd/models.py
cjmabry: Agreed.

bsd/models.py
```diff
@@ -43,9 +43,10 @@ def assert_valid_account(api_result):
     assert cons.findtext('is_banned') == "0"
 
 
-def find_constituents_by_state_cd(state_cd):
+def find_constituents_by_state_cd(state_cd, cons_group=None):
```
ericbroder: would be good to update params doc for method too. seems like it should be a string id or a list of string ids, right?

ericbroder: would be good to update params doc for method too. seems like it should be a string id or a list of string ids, right?

bsd/models.py
```diff
@@ -64,6 +65,11 @@ def find_constituents_by_state_cd(state_cd):
     filter = {}
     filter['state_cd'] = str(state_cd)
     filter['is_subscribed'] = True
+
+    """Check if member of sendable cons group"""
```
ericbroder: comment is out of date. just check if cons group param is present and add to filter if not None, or something like that

organizing_hub/tasks.py
```diff
@@ -238,6 +239,9 @@ def build_contact_list_for_event_promotion(event_promotion_id):
     event_promotion_id : int
         EventPromotion id
 
+    cons_group : int
+        BSD constituent group id
```
ericbroder: there is no param in this method for cons group?

cjmabry: Agh wrong method.

bsd/models.py
```diff
@@ -54,6 +55,9 @@ def find_constituents_by_state_cd(state_cd):
     state_cd : str
         BSD field for state/territory code, 2 characters
 
+    cons_group : int
```
ericbroder: needs to be a string or list of strings when passed to bsd api filter class

ericbroder: also you are using a string format already here, aside from this comment

PR #335, opened 2018-09-19
 TECH-1397 dont send promotion if event is in the past or not approved

cjmabry:


organizing_hub/signals.py
```diff
@@ -128,17 +128,19 @@ def event_promotion_post_save_handler(instance, **kwargs):
...
+            list_name = 'List for Event Promotion: ' + str(event_promotion)
+            contact_list = ContactList.objects.create(name=list_name)
+            event_promotion.contact_list = contact_list
+            event_promotion.save()
+
+            """Call async task to build list"""
+            build_contact_list_for_event_promotion.delay(event_promotion.id)
+        else:
+            """Resave contact list to trigger post_save logic"""
+            contact_list.save()
```
ericbroder: Why did you add this `else` clause? I don't understand the purpose. Why do we need to save the contact list if promo is approved and contact list is not `None`?

organizing_hub/tasks.py
```diff
@@ -321,6 +321,20 @@ def send_event_promotion(event_promotion_id):
     if event_promotion.status != EventPromotionStatus.approved.value[0]:
         return sent_count
 
+    """If event is not approved, then do nothing"""
+    event = find_event_by_id_obfuscated(event_promotion.event_external_id)
+
+    if event.status != 'Approved':
```
ericbroder: Need to add `None` handling if no event is returned.

ericbroder: Also ideally would match on status id or key rather than human readable string. This was initially just meant for display content.

organizing_hub/signals.py
ericbroder: I'm worried this may have side effects. As written previously, it would only do something if promo was approved with no contact list. Now it does stuff if that is the case, or if contact list is already there. But if contact list is not `None`, then it should already have started to build list and then send emails, so why are we triggering anything additional in this case?

cjmabry: When an event promo is marked as event not approved, an admin might go and approve the event, and then come back and mark the event promo as approved. This should trigger send again. As-is, the first time an event is marked as approved a list is built, but the send method will not send based on many conditions (such as if the event isn't approved). Some of these conditions aren't permanent, however, like event not approved. Perhaps instead of following precedent and returning from send event function if conditions don't pass we should instead not build list until promo is ready to go out.

ericbroder: @cjmabry the prior version of logic was meant to avoid accidentally resending to same list twice, or sending to a stale list. For admin retry send, the process outlined in https://ourrevolution.atlassian.net/browse/TECH-1344 is to set contact list to None and re-approve the promotion. This ensures that a fresh list is regenerated.

ericbroder: The other issue with the change you suggested is that it's confusing. It saves contact list for no clear reason. So even with that route, would be better to add more code comments, or call the method needed directly rather than indirectly via a post save trigger.

bsd/models.py
```diff
@@ -385,6 +386,22 @@ def _get_absolute_url(self):
...
+            '%Y-%m-%d %H:%M:%S'
+        )
+        local_datetime = event_time_zone.localize(naive_datetime)
+        utc_datetime = local_datetime.astimezone(pytz.utc)
+        return utc_datetime
+    start_datetime_utc = property(_get_start_datetime_utc)
+
+    def _is_approved(self):
+        if not self.flag_approval == 1:
+            return True
```
ericbroder: `return not self.flag_approval == 1`?

ericbroder: Also wonder if we can just remove/replace the other `status` property now that we have this? I like this new implementation, although now the other thing seems redundant.

events/models.py
```diff
@@ -50,6 +50,8 @@ class EventPromotionStatus(Enum):
     in_progress = (20, 'In Progress')
     sent = (30, 'Sent')
     skipped = (40, 'Skipped')
+    event_not_approved = (50, 'Event Not Approved')
+    expired = (60, 'Expired')
```
ericbroder: realizing that we made explicit statuses for a couple reasons that event promo will fail, but didn't make statuses for several other reasons it would fail. kind of inconsistent. not sure what is best 🤔

organizing_hub/tasks.py
```diff
@@ -321,6 +321,24 @@ def send_event_promotion(event_promotion_id):
...
+            event_promotion.status = EventPromotionStatus.event_not_approved.value[0]
+            event_promotion.save()
+            return sent_count
+
+        """If event is in past, then do nothing"""
+        if event.start_datetime_utc <= timezone.now():
+            event_promotion.status = EventPromotionStatus.expired.value[0]
+            event_promotion.save()
+            return sent_count
+    else:
```
ericbroder: seems like this would be good case for `if ... elif ... elif` pattern. minor issue, just would read a little easier.

events/models.py
ericbroder: i guess its good to have statuses for common fail scenarios that are caused by user error, like event not approved or in the past. and for fail scenarios that are rare or theoretically should never happen i guess we don't need explicit statuses.

another option would be to validate the promotion save and disallow setting promotion to approved for event in past or event not approved 🤔

organizing_hub/signals.py
```diff
@@ -129,17 +129,20 @@ def event_promotion_post_save_handler(instance, **kwargs):
...
-        contact_list = ContactList.objects.create(name=list_name)
-        event_promotion.contact_list = contact_list
+            """Create new contact list and add to event promotion"""
+            list_name = 'List for Event Promotion: ' + str(event_promotion)
+            contact_list = ContactList.objects.create(name=list_name)
+            event_promotion.contact_list = contact_list
+            event_promotion.save()
+
+            """Call async task to build list"""
+            build_contact_list_for_event_promotion.delay(event_promotion.id)
```
ericbroder: indentation is wrong? which i think in python might actually matter

events/models.py
```diff
@@ -50,7 +50,15 @@ class EventPromotionStatus(Enum):
...
 
+"""Statuses that require clearing of the contact list"""
+event_promotion_statuses_for_list_clear = [
```
ericbroder: I wonder if this logic makes more sense in organizing hub app rather than here. Kind of related to specific event promo implementation and signal/task logic. Not necessarily universal to all event promo models 🤔

events/models.py
```diff
@@ -84,6 +92,13 @@ class EventPromotion(models.Model):
     """Assume generic external user id for event owner/host"""
     user_external_id = models.CharField(db_index=True, max_length=128)
 
+    def _requires_list_clear(self):
```
ericbroder: ditto seems like maybe this should be in hub app rather than on the model. also it's a bit confusing when list was already clear, this will still say it requires clear. maybe check if list is already cleared or not?

events/models.py
```diff
@@ -84,6 +92,13 @@ class EventPromotion(models.Model):
...
+        """Check if status is in list of statuses that requires a list clear"""
+        return self.status in [
+            x.value[0] for x in event_promotion_statuses_for_list_clear
```
ericbroder: nice 👍

organizing_hub/signals.py
```diff
@@ -129,17 +129,20 @@ def event_promotion_post_save_handler(instance, **kwargs):
...
+
+    """Clear the contact list if it requires clearing"""
+    if event_promotion.requires_list_clear and contact_list is not None:
```
ericbroder: `elif` is slightly better here, not a big deal.

PR #344, opened 2018-10-03
TECH-1415 Signify nominations closing for 2018

cjmabry:


pages/templates/partials/past_results.html
```diff
@@ -36,31 +36,25 @@ <h2 class="mb20">
           </a>
           {% endif %}
 
-          <hr>
+          {% comment %}Wagtail places HTML tags in blank RichTextFields, so a
```
ericbroder: Thanks for added the code comments to explain this. Another option is to add a boolean checkbox to toggle display of content.

pages/templates/partials/past_results.html
```diff
@@ -36,31 +36,25 @@ <h2 class="mb20">
...
-            </strong>
-          </p>
+          See: https://github.com/wagtail/wagtail/issues/2336
+          {% endcomment %}
+          {% if page.secondary_copy|striptags %}
+          <hr>
+          {{ page.secondary_copy|richtext }}
 
+          {% if page.button_show %}
           <a href="/groups/nominations"
```
ericbroder: Might be out of scope but should use template tag or something like that rather than hardcode url into template.

pages/models.py
```diff
@@ -818,9 +818,26 @@ def candidate_endorsement_unpublished_handler(instance, **kwargs):
...
     content_heading = models.CharField(max_length=128, blank=True, null=True)
+    secondary_copy = RichTextField(
+        blank=True,
+        null=True,
+        help_text='Copy to go below Past Election Results section.'
+    )
+    button_show = models.BooleanField(
+        default=False,
+        help_text="""Show nominations platform Get Started button. Will only
+        display if secondary copy is present."""
```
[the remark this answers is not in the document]
cjmabry: I don't want to allow political team to accidentally leave just a Get Started button on there. The button is not clear without accompanying copy.

nominations/templates/email/candidate_email.html
```diff
@@ -6,21 +6,23 @@ <h1 style="text-align:center;padding:25px 0 20px 0;border:0;margin:0">
...
 
-  <p style="font-size: 17px; font-family: Arial; line-height:140%;">You’ve been nominated by {{group}} for an endorsement from Our Revolution!</p>
+  <p style="font-size: 17px; font-family: Arial; line-height:140%;">You’ve been nominated by {{group}} for an endorsement from Our Revolution! Your local endorsement is a critical first step in building the Political Revolution across the land.</p>
 
-  <p style="font-size: 17px; font-family: Arial; line-height:140%;">Congratulations on receiving an endorsement nomination from a local Our Revolution group! In order for Our Revolution to consider your candidacy for a national endorsement, you must fill out and submit the following questionnaire.</p>
+  <p style="font-size: 17px; font-family: Arial; line-height:140%;"><strong>We have closed nominations for the 2018 election season.</strong> If your election is for 2019, please feel free to fill out the questionnaire you will receive. If your election day is this year, there is no need to follow up. We have informed the group that nominated you that we are no longer accepting nominations for this year. We wish you luck for this year regardless.</p>
```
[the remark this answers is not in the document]
cjmabry: Yes, didn't want to change content but this makes sense for our purposes.

nominations/templates/index.html
```diff
@@ -12,6 +12,8 @@
   <div class="row blue-bg nina-bg white-text pb4">
 
     <div class="container pt2">
+      {% nominations_platform_alert %}
```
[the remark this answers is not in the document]
cjmabry: 👍

nominations/templates/email/candidate_email.html
```diff
@@ -6,21 +6,23 @@ <h1 style="text-align:center;padding:25px 0 20px 0;border:0;margin:0">
...
+  <p style="font-size: 17px; font-family: Arial; line-height:140%;"><strong>Please keep in mind that the following factors will be essential for consideration of a national endorsement:</strong></p>
 
   <p style="font-size: 17px; font-family: Arial; line-height:140%;"><strong>1. Values:</strong> How aligned is the candidate with Our Revolution’s values?</p>
 
-  <p style="font-size: 17px; font-family: Arial; line-height:140%;"><strong>2. Viability:</strong> Historical voting data, including voter turnout and previous election party breakdown, is a good indicator of how an election may turn out. This information will give an indication of how much effort a candidate will need to exert and invest in order to win.</p>
+  <p style="font-size: 17px; font-family: Arial; line-height:140%;"><strong>2. Pathway:</strong> Our Revolution does not expect all our candidates will win - sometimes we know they won’t. But we want candidates to demonstrate that they have thought about how they would win. This thought exercise is called “a pathway to victory,” we want you to think about what steps you can take to win the race by reaching out to voters and allies.</p>
```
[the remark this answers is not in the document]
cjmabry: Not sure the change makes it any clearer, I still read it the same way. This was political's final copy so really up to them.

ourrevolution/settings.py
```diff
@@ -16,6 +16,12 @@
 
 mimetypes.add_type('image/svg+xml', 'svg')
 
+ALERT_LEVELS = (
```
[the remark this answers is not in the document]
cjmabry: I was going to go enum but was hitting end of ticket. Will see if I can revisit.

nominations/views.py
```diff
@@ -402,6 +402,8 @@ def send_notification(self, application):
 
         d = {
             'or_logo_secondary': settings.OR_LOGO_SECONDARY,
+            'group': group_name,
```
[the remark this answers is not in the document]
cjmabry: 👍

nominations/models.py
```diff
@@ -90,9 +91,23 @@ class NominationsPlatformAlert(models.Model):
         help_text='Show alert on nominations platform pages.'
     )
 
+    alert_level = models.CharField(
+        max_length=16,
+        choices=[x.value for x in AlertLevels],
+        default=AlertLevels.warning.value[0],
```
ericbroder: Much more readable 👍

nominations/models.py
```diff
@@ -90,9 +91,23 @@ class NominationsPlatformAlert(models.Model):
         help_text='Show alert on nominations platform pages.'
     )
 
+    alert_level = models.CharField(
+        max_length=16,
+        choices=[x.value for x in AlertLevels],
```
ericbroder: Nice 👍

nominations/models.py
```diff
@@ -90,9 +91,23 @@ class NominationsPlatformAlert(models.Model):
         help_text='Show alert on nominations platform pages.'
     )
 
+    alert_level = models.CharField(
```
ericbroder: Suggest to make this an integer field, to decouple the id we store in db from the actual css class that goes into the template.

organizing_hub/models.py
```diff
@@ -37,6 +37,12 @@ def serve(self, request, *args, **kwargs):
             **kwargs
         )
 
+@unique
+class AlertLevels(Enum):
```
ericbroder: This is much better 😂 If possible I suggest to use integer ids though, to decouple css class from db id.

Also this model is really for all bootstrap alerts, not just organizing hub, so I'm not sure exactly which app it belongs in?

nominations/models.py
```diff
@@ -91,9 +91,10 @@ class NominationsPlatformAlert(models.Model):
         help_text='Show alert on nominations platform pages.'
     )
 
-    alert_level = models.IntegerField(
-        choices=ALERT_LEVELS,
-        default=3,
+    alert_level = models.CharField(
+        max_length=16,
```
ericbroder: Suggest to add code comment or used a named variable etc., otherwise this is a confusing "magic number" that seems to come from nowhere.

nominations/templates/index.html
```diff
@@ -12,7 +12,9 @@
   <div class="row blue-bg nina-bg white-text pb4">
 
     <div class="container pt2">
-      {% nominations_platform_alert %}
+      <div class="col-md-12">
```
ericbroder: There should always be a row in between container and column. Also `col-md-12` is the same as `col-xs-12` fyi.

nominations/views.py
```diff
@@ -278,7 +278,7 @@ def form_valid(self, form):
 
         candidate_name = application.candidate_first_name + ' ' + application.candidate_last_name
         candidate_email = form.cleaned_data['candidate_email']
-        group = application.group
+        group_name = application.group
```
ericbroder: Isn't this a local group object? Should be `group_name  = application.group.name` or something like that?

nominations/views.py
```diff
@@ -402,7 +402,7 @@ def send_notification(self, application):
 
         d = {
             'or_logo_secondary': settings.OR_LOGO_SECONDARY,
-            'group': group_name,
+            'group_name': group_name,
```
ericbroder: nice 👍

organizing_hub/models.py
cjmabry: I was going to put it in pages but not sure if that makes sense either. So had to pick one.

nominations/models.py
cjmabry: I don't know if that makes the most sense with use of enums, because we need to get the value. The value in this case is the css class. If it's an integer we'll have to do custom logic on the template side to translate the integers to a css value. Thoughts?

nominations/templates/email/candidate_email.html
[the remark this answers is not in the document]
ericbroder: my suggestion is that `sometimes we know they won’t ` sounds to me like sometimes we know the result will be a loss before the election takes place, which is odd. but `we know sometimes they won’t` sounds to me like we know  that probabilistically some of the candidates will lose, which is true.

nominations/models.py
cjmabry: 👍

nominations/templates/index.html
cjmabry: 👍

nominations/views.py
cjmabry: 👍

nominations/models.py
ericbroder: We already have examples with more than 2 values per tuple in the Enum. You can do something like enum_member = (integer id, human display name, css class). See `class CallQuestion(Enum)`. And then the field looks like ```
    question = models.IntegerField(choices=[
        (x.value[0], x.value[1]) for x in CallQuestion
    ])
```

File: organizing_hub/models.py (hunk 12 above)
Comment (ericbroder) [reply organizing_hub/models.py]: I thought about pages app too, but that's just for wagtail stuff, and this could be used basically anywhere we have a bootstrap alert, so basically anywhere on the site. It's almost like it should be part of a bootstrap app, or a front end assets app.

Comment (ericbroder) [reply organizing_hub/models.py]: interesting: https://django-bootstrap3.readthedocs.io/en/latest/templatetags.html#bootstrap-alert

Comment (ericbroder) [reply organizing_hub/models.py]: yup here it is: https://github.com/dyve/django-bootstrap3/blob/8.2.3/bootstrap3/templatetags/bootstrap3.py#L26
```
MESSAGE_LEVEL_CLASSES = {
    DEFAULT_MESSAGE_LEVELS.DEBUG: "alert alert-warning",
    DEFAULT_MESSAGE_LEVELS.INFO: "alert alert-info",
    DEFAULT_MESSAGE_LEVELS.SUCCESS: "alert alert-success",
    DEFAULT_MESSAGE_LEVELS.WARNING: "alert alert-warning",
    DEFAULT_MESSAGE_LEVELS.ERROR: "alert alert-danger",
}
```

Comment (ericbroder) [reply organizing_hub/models.py]: https://github.com/dyve/django-bootstrap3/blob/8.2.3/bootstrap3/components.py#L27

File: nominations/templates/index.html (hunk 14 above)
Comment (cjmabry) [reply nominations/templates/index.html]: Actually rows aren't required, and I'd rather not rewrite this whole page to add rows.

Comment (cjmabry) [reply nominations/templates/index.html]: Or, rewording that, fixing this page to be proper rows and container structure is outside of scope here.

Comment (ericbroder) [reply nominations/templates/index.html]: I'm not sure what you mean. You don't need to fix the whole page. You just need to put `<div class="row">` in between the container and col divs. It should be a trivial change?

Bootstrap grid apparently supports using container without rows or columns. However for the grid it should be container > row > column. So can we please do one or the other?

Also would be nice to have conditional None handling if there is no alert, but that's probably out of scope.

File: organizing_hub/models.py (hunk 12 above)
Comment (cjmabry) [reply organizing_hub/models.py]: Hmm, could reimplement the alerts to use render alert here but still need the enum to get choices from for CMS - importing those would be pretty brittle I think.

Comment (cjmabry) [reply organizing_hub/models.py]: I think for now it makes most sense in pages, creating a new front-end app is way outside scope here.

File: nominations/templates/index.html (hunk 14 above)
Comment (cjmabry) [reply nominations/templates/index.html]: Since the rest of the page doesn't have rows here we need to do that on every container and column on this page or we'll have different padding in different sections. On top of that we have nested containers which aren't allowed in bootstrap so really we need a full reworking here. This is in line with how page has been built and isn't wrong per se so don't think it's worth spending time here fixing when already over on this ticket.

File: organizing_hub/models.py (hunk 12 above)
Comment (cjmabry) [reply organizing_hub/models.py]: I'm not sure how this is making things simpler over just using a config similar to how django bootstrap implemented? What is enum trying to solve here? Seems to be making things much more complex.

Comment (ericbroder) [reply organizing_hub/models.py]: I'm not asking you to refactor this right now to use the django bootstrap library. I'm just pointing out that it is one way to do it. The most important thing at the moment I think is that this should be an Enum and not a setting.

Personally I think putting all our Enums into a single settings file is the more complex option. Settings is not a catchall for any constant/variable we have in the app which is not stored in the db. It should be used for things which actually fit the idea of a config setting, so we can have proper separation of layers.

Django bootstrap is not using the settings file, it is using template tags. It has a slightly different use case and they decided to use a dictionary instead of Enum, I don't know why.

File: nominations/templates/index.html (hunk 14 above)
Comment (cjmabry) [reply nominations/templates/index.html]: 👍

File: pages/templatetags/pages_tags.py [hunk 17]
```diff
@@ -27,6 +27,17 @@ def candidates_url():
 def endorsement_process_url():
     return settings.ENDORSEMENT_PROCESS_URL
 
+@register.simple_tag
+def get_alert_level_class(value):
+    """Pass in alert level value and get back approporiate CSS class."""
+    classes_dict = {
+        1: 'success',
+        2: 'info',
+        3: 'warning',
```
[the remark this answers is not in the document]
cjmabry: Good call. For second point I'm going to leave just as `info` `warning` etc. because they can be used with other bootstrap classes should we ever need them, not just alerts.

pages/templatetags/pages_tags.py
```diff
@@ -28,6 +28,18 @@ def endorsement_process_url():
...
 
 
+@register.simple_tag
+def get_alert_level_class(value):
+    """Pass in alert level value and get back approporiate CSS class."""
+    classes_dict = {
+        AlertLevels.success.value[0]: 'success',
+        AlertLevels.info.value[0]: 'info',
+        AlertLevels.warning.value[0]: 'warning',
+        AlertLevels.danger.value[0]: 'danger'
```
ericbroder: 😂

PR #345, opened 2018-10-03
TECH-1425 Simplify name on ballot initiatives

cjmabry:


pages/models.py
```diff
@@ -988,6 +988,19 @@ class InitiativeEndorsementPage(Page):
         ImageChooserPanel('social_image')
     ]
 
+    def _get_initiative_title(self):
+        """Build title for initiative and support legacy pages"""
+        if self.initiative_title and self.initiative_name:
```
ericbroder: What if one of these is blank/None?

pages/models.py
```diff
@@ -988,6 +988,19 @@ class InitiativeEndorsementPage(Page):
...
+            initiative_title = (
+                str("Yes" if self.how_to_vote else "No") + ' on ' + self.initiative_title
+                + ': ' + self.initiative_name
+            )
+        else:
+            initiative_title = self.title
+
+        return initiative_title
+    get_initiative_title = property(_get_initiative_title)
```
ericbroder: We don't really have a standard here but I usually use `get____` for methods and just a plain non-prefixed name for properties/fields. Like make a property called `display_title` or something like that?

This is also confusing because we have a field on the model called `initiative_title` but it's different from `get_initiative_title`. Can we make it more distinct name so they don't get mixed up?

pages/templates/partials/initiative_endorsement_old.html
```diff
@@ -10,7 +10,7 @@
             <div class="row">
                 <div class="container">
                     <div class="page-sub-nav">
-                        <a href="{% pageurl page.get_parent %}"><span class="glyphicon glyphicon-chevron-left"></span></a> Our Initiatives / {{ initiative.state }} / {{ initiative.title }}: {{ initiative.name }}
+                        <a href="{% pageurl page.get_parent %}"><span class="glyphicon glyphicon-chevron-left"></span></a> Our Initiatives / {{ initiative.state }} / {{ initiative.vote | yesno | title }} on {{ initiative.title }}: {{ initiative.name }}
```
ericbroder: why not use the new property you created?

pages/templates/partials/initiative_snippet_result.html
```diff
@@ -13,7 +13,7 @@
 
           <div class="card__content full initiative">
             <div>
-              <span class="card__item name bold tight"><span>{{ initiative.title }}: {{ initiative.name }}</span></span>
+              <span class="card__item name bold tight"><span>{{ initiative.vote|yesno|title }} on {{ initiative.title }}: {{ initiative.name }}</span></span>
```
ericbroder: ditto, use new property?

pages/models.py
cjmabry: It'll just default to title which is suitable.

pages/models.py
cjmabry: 👍

pages/templates/partials/initiative_snippet_result.html
cjmabry: These don't all have pages attached and it's a bit confusing what will have what so I just left as-is here because it already supports legacy. I just added vote to be more consistent across pages.

pages/templates/partials/initiative_endorsement_old.html
cjmabry: 👍

PR #353, opened 2018-10-09
TECH-1442 TECH-1452 generate list for call campaign and fix location bug

ericbroder:


bsd/models.py
```diff
@@ -52,6 +52,7 @@ def find_constituents_by_state_cd(
     subscribers_only=True,
     primary_address_only=True,
     with_email=False,
+    with_phone=False,
```
cjmabry: Nice, good find.

PR #358, opened 2018-10-16
TECH-1387 add feature flags for organizing hub

ericbroder:


organizing_hub/mixins.py
```diff
@@ -39,6 +42,28 @@ def handle_no_permission(self):
         )
         return redirect(settings.ORGANIZING_HUB_DASHBOARD_URL)
 
+    def has_feature_access(self):
+        """Return True if we are skiping feature check"""
```
cjmabry: Small typo on skipping.

```suggestion
        """Return True if we are skipping feature check"""
```

organizing_hub/mixins.py
```diff
@@ -39,6 +42,28 @@ def handle_no_permission(self):
...
+        if self.skip_feature_check:
+            return True
+
+        """Throw error is feature is missing"""
```
cjmabry:
```suggestion
        """Throw error if feature is missing"""
```

PR #360, opened 2018-10-17
TECH-1389 Add callers to create calling campaign page and implement edit campaign page

cjmabry:


organizing_hub/views/call_views.py
```diff
@@ -86,6 +91,34 @@ class CallCampaignDetailView(LocalGroupPermissionRequiredMixin, DetailView):
...
+    Your calling campaign has been edited succesfully.
+    '''
+
+    def get_context_data(self, **kwargs):
+        context = super(CallCampaignUpdateView, self).get_context_data(**kwargs)
+        context['update_view'] = True
+        return context
+
+    def get_local_group(self):
+        return find_local_group_by_user(self.request.user)
```
ericbroder: This needs to get local group from campaign for mixin, not from user.

calls/forms.py
```diff
@@ -32,9 +92,14 @@ class Meta:
         ]
         model = CallCampaign
         widgets = {
-            'script': forms.Textarea(attrs={'rows': '14'}),
+            'script': forms.Textarea(attrs={'rows': '14'})
         }
 
+class CallCampaignUpdate(CallCampaignForm):
```
ericbroder: add suffix `_Form`

calls/forms.py
```diff
@@ -1,13 +1,72 @@
+from bsd.models import BSDProfile
```
ericbroder: calls app should not import anything from BSD. organizing hub imports both calls and bsd, but calls app should not import bsd.

calls/forms.py
```diff
@@ -1,13 +1,72 @@
...
+from django.forms.utils import flatatt
+from django.utils.html import mark_safe
+from .models import CallCampaign, CallProfile
 
 CALLS_MAX_DISTANCE_MILES = settings.CALLS_MAX_DISTANCE_MILES
 CALLS_MAX_LIST_SIZE = settings.CALLS_MAX_LIST_SIZE
 
+class CommaSeparatedCallersTextArea(Widget):
+    def render(self, name, value, attrs=None, renderer=None):
+        final_attrs = self.build_attrs(attrs, {'type':'text', 'name':name})
```
ericbroder: is `'name':name` correct? i dont really understand what this does but seems like an odd value to put for name of something.

calls/forms.py
```diff
@@ -1,13 +1,72 @@
...
+        caller_emails = []
+
+        if value is not None:
+            for caller_id in value:
```
ericbroder: please add code comments to explain what is happening here.

calls/forms.py
```diff
@@ -1,13 +1,72 @@
...
+        attrs={
+            'rows': '5',
+            'placeholder': 'bob@example.com, sally@example.comm, tom@example.com'
+        }
+    )
+
+    def clean(self, value):
+        caller_ids = []
+
+        if value is not None and value != '':
```
ericbroder: Please add code comments to explain custom logic.

calls/forms.py
```diff
@@ -1,13 +1,72 @@
...
+            value = [email.strip() for email in value.split(",")]
+
+            for email in value:
+                # TODO: support multiple accounts per email
+                if User.objects.filter(email__iexact=email).exists():
+                    user = User.objects.get(email__iexact=email)
+                else:
+                    user = User.objects.create_user(
```
ericbroder: This is not form validation, this is complex database logic. I don't think this is what the form `clean` method is meant for. If you look at the `GroupAdminsView` example, all of the db object creation is happening in the view, not the form. I suggest moving this logic to the view, under `form_valid` or something like that.

calls/forms.py
```diff
@@ -1,13 +1,72 @@
...
+    def clean(self, value):
+        caller_ids = []
+
+        if value is not None and value != '':
+            value = [email.strip() for email in value.split(",")]
+
+            for email in value:
+                # TODO: support multiple accounts per email
+                if User.objects.filter(email__iexact=email).exists():
+                    user = User.objects.get(email__iexact=email)
```
ericbroder: This is doing 2 different db queries to lookup the same object twice. Please simply this to just one query.

Might also want to use `get_or_create` method: https://docs.djangoproject.com/en/1.11/ref/models/querysets/#get-or-create

organizing_hub/templates/calls/callcampaign_detail.html
```diff
@@ -124,18 +127,29 @@ <h3>Callers</h3>
...
-        <a href="todo">
-          View
+        <a href="{% url 'organizing-hub-call-campaign-update' campaign.uuid %}">
+          Edit
         </a>
       </p>
 
+      <p>
+        Callers will be able to login and make calls using the email address
+        listed after creating an account. You may add or remove callers at any
```
ericbroder: I was a bit confused by language `Callers will be able to login and make calls using the email address listed after creating an account.` Suggest something more like `Callers will be able to login and make calls after creating an account using the email address specified here.`

organizing_hub/templates/calls/callcampaign_form.html
```diff
@@ -10,7 +10,7 @@
 {% endblock %}
 
 {% block content_heading %}
-Create Campaign
+{% if update_view %}Update{% else %}Create{% endif %} Campaign
```
ericbroder: Could also put campaign title here for update?

organizing_hub/templates/calls/callcampaign_form.html
```diff
@@ -67,6 +70,19 @@ <h3>Contact Info</h3>
 
         </div>
 
+        <h3>Callers</h3>
+        <p>
+          Add callers to your calling campaign. These users will be able to
+          login and make calls using the email address listed after creating an
+          account. You may add or remove callers after campaign creation.
```
ericbroder: Ditto on language. Also if we're going to reuse content then please put into a central partial or message variable.

organizing_hub/urls.py
```diff
@@ -148,6 +149,11 @@
                         CallCampaignDetailView.as_view(),
                         name='organizing-hub-call-campaign-detail'
                     ),
+                    url(
+                        r'^(?P<uuid>[0-9a-f-]+)/update/',
```
ericbroder: There is now a parent level /uuid/ url and anything under that path should be nested inside that url declaration. see call view and detail view for example. Generally don't want to redeclare url paths if we can avoid repeating ourselves.

organizing_hub/views/call_views.py
```diff
@@ -86,6 +91,35 @@ class CallCampaignDetailView(LocalGroupPermissionRequiredMixin, DetailView):
...
 
+class CallCampaignUpdateView(
+    LocalGroupPermissionRequiredMixin,
+    SuccessMessageMixin,
+    UpdateView
+):
+    template_name = "calls/callcampaign_form.html"
+    form_class = CallCampaignUpdateForm
+    model = CallCampaign
+    permission_required = 'calls.change_callcampaign'
```
ericbroder: Merge issue, this needs a feature flag now.

organizing_hub/templates/calls/callcampaign_form.html
ericbroder: Ok nevermind, I see now there is an editable title field.

organizing_hub/urls.py
cjmabry: 👍

organizing_hub/views/call_views.py
cjmabry: 👍

organizing_hub/templates/calls/callcampaign_form.html
cjmabry: 👍

calls/forms.py
cjmabry: 👍

calls/forms.py
cjmabry: 👍

calls/forms.py
cjmabry: 👍

calls/forms.py
cjmabry: 👍

calls/forms.py
cjmabry: :+1:

organizing_hub/templates/calls/partials/call_campaign_callers_explainer.html
```diff
@@ -0,0 +1,4 @@
+<p>
```
cjmabry: I didn't wanna put this in a config because it seemed weird but not sure if this makes the most sense either.

organizing_hub/views/call_views.py
```diff
@@ -121,6 +129,46 @@ def can_make_call_for_campaign(user, call_campaign):
...
+        for email in caller_emails:
+            # TODO: support multiple accounts per email
+            (user, created) = User.objects.get_or_create(
+                email=email,
+                defaults={
+                    'username':email
+                }
+            )
+
+            if not hasattr(user,'bsdprofile'):
```
cjmabry: I still need to move the BSD Profile creation out, or this entire function out. Thoughts on where to put this or how to structure? Perhaps a function that handles this in organizing hub that's imported here?

calls/forms.py
```diff
@@ -14,6 +19,27 @@
...
+    """
+    def render(self, name, value, attrs=None, renderer=None):
+        final_attrs = self.build_attrs(attrs, {'type':'text', 'name':name})
+        caller_emails = []
+
+        if value is not None:
+            """Loop through the caller_ids we have stored on the many-to-many
+            relationship and get the corresponding object"""
+            for caller_id in value:
+                callprofile = CallProfile.objects.get(pk=caller_id)
```
ericbroder: I don't think a form field widget render function should be creating new records in db. Doesn't this violate the purpose of the render method?
```
render(name, value, attrs=None, renderer=None)[source]¶
Renders a widget to HTML using the given renderer. If renderer is None, the renderer from the FORM_RENDERER setting is used.
```
https://docs.djangoproject.com/en/1.11/ref/forms/widgets/#django.forms.Widget.render

Each piece of the software app has a specific purpose. Some are for creating db records. Some are for rendering html. Please think about what type of logic is happening and where it should be located based on the overall design pattern and separation of concerns principle: https://en.wikipedia.org/wiki/Separation_of_concerns

calls/forms.py
```diff
@@ -14,6 +19,27 @@
...
+                user = callprofile.user
+                caller_emails.append(str(user.email))
+
+            """Build list of comma seprated values for display"""
```
ericbroder: typo *separated

organizing_hub/templates/calls/partials/call_campaign_callers_explainer.html
ericbroder: Generally I think this is ok. Ideally all of our content strings would be in some sort of messages list, and not hardcoded into templates. This way we can separate content from template logic. Also helps for translation/multi-language support. Then we would pull content from the messages list by a key.

However I do see that you are including `p` tags inside the partial, but in one area you also have `p` tags around the call to partial, and another area you don't. So need to be consistent here.

organizing_hub/views/call_views.py
```diff
@@ -121,6 +129,46 @@ def can_make_call_for_campaign(user, call_campaign):
...
+                BSDProfile.objects.create(user=user)
```
ericbroder: Might need to be cautious about django user vs bsd user, not sure if it matters here.

organizing_hub/views/call_views.py
ericbroder: Why do you need to move it out? Previously I asked to move it because it was in the calls app form layer. Now it's in the organizing hub app views layer, which is much better.

cjmabry: Wait I had forgotten I moved it here, disregard actually.

ericbroder: Although in general I think we should explore some sort of service layer which is not models and not views. I have been putting service api methods in models or views, but really would be nice to have some sort of service api layer in these apps. Not sure what is best practice in django.

organizing_hub/views/call_views.py
```diff
@@ -292,6 +340,17 @@ def form_valid(self, form):
...
         form.instance.owner = self.request.user.callprofile
+
+        """Take comma separated caller emails and get or create corresponding
+        call profiles"""
+        caller_emails = form.cleaned_data['caller_emails']
+
+        """Save call_campaign because we need a primary key in order to save
+        Many-to-many caller relationships"""
+        call_campaign = form.save()
+        caller_ids = get_or_create_callers(caller_emails)
```
ericbroder: Maybe better to put this in a post save signal?

organizing_hub/views/call_views.py
ericbroder: Maybe move to post-save signal?

calls/forms.py
ericbroder: I'm not sure if there should be any db queries happening in a form widget render method, but at least it's just a get and not data change.

calls/forms.py
cjmabry: Can you clarify here? Name is an html attribute that django uses to populate data to correct field.

organizing_hub/views/call_views.py
cjmabry: I don't think it makes sense to put in post_save sense it only relates to this form really. Wouldn't want to do this logic after an admin form save, for example.

calls/forms.py
cjmabry: I agree, I'm going to try and write out the widget.

calls/forms.py
cjmabry: 👍

organizing_hub/templates/calls/partials/call_campaign_callers_explainer.html
cjmabry: 👍

calls/forms.py
```diff
@@ -1,11 +1,10 @@
 from django import forms
 from django.conf import settings
-from django.forms import widgets
 from contacts.models import ContactListStatus
 from .models import (
     call_campaign_statuses_skip_list_validation,
     CallCampaign,
-    CallQuestion,
+    CallQuestion
```
ericbroder: we dont really have a standard here, but i have been using the trailing comma on lists like this, which is acceptable in python. and you apparently do not. so we should agree at some point rather than keep changing each other's commas.

calls/forms.py
```diff
@@ -90,9 +94,14 @@ class Meta:
         ]
         model = CallCampaign
         widgets = {
-            'script': forms.Textarea(attrs={'rows': '14'}),
+            'script': forms.Textarea(attrs={'rows': '14'})
```
ericbroder: 😆

organizing_hub/templates/calls/callcampaign_detail.html
```diff
@@ -129,7 +129,10 @@ <h3 class="mt0">Campaign Info</h3>
         </div>
       </div>
 
-      <a class="pull-right" href="todo">
+      <a
+        class="pull-right fs16"
```
ericbroder: suggest to use the more semantic name like `fs-lg` etc. minor issue.

organizing_hub/templates/calls/callcampaign_detail.html
```diff
@@ -141,11 +144,21 @@ <h3>Callers</h3>
...
-        <a href="todo">
-          View
+        <a href="{% url 'organizing-hub-call-campaign-update' campaign.uuid %}">
+          Edit
         </a>
       </p>
 
+      {% include "calls/partials/call_campaign_callers_explainer.html" %}
+
+      {% if campaign.callers.all|length > 0 %}
```
ericbroder: do we need empty list logic?

organizing_hub/templates/calls/callcampaign_detail.html
```diff
@@ -141,11 +144,21 @@ <h3>Callers</h3>
...
+      <div class="mb10 well well-sm">
+        {% for caller in campaign.callers.all  %}
+        {{ caller.user.email }}{% if not forloop.last %}, {% endif %}
```
ericbroder: `{% if not forloop.last %}` interesting 🤔

organizing_hub/templates/calls/callcampaign_form.html
```diff
@@ -33,9 +33,12 @@
 
     <div class="col-md-9">
 
-      <form action="{% url 'organizing-hub-call-campaign-create' %}"
-      method="post" class="form mb20">
-
+      {% if update_view %}
+      {% url 'organizing-hub-call-campaign-update' object.uuid as action_url %}
```
ericbroder: Suggest to be consistent, either use `campaign ` or `object` in the template. I like the more semantic version, rather than `object`.

organizing_hub/templates/calls/callcampaign_form.html
```diff
@@ -67,6 +70,20 @@ <h3>Contact Info</h3>
...
+          after campaign creation.
+        </p>
+
+        {% include "calls/partials/call_campaign_callers_explainer.html" %}
+
+        <p>
+          Enter one or more email addresses, separated by commas.
+        </p>
+
+        {% bootstrap_field form.caller_emails placeholder="joe@example.com, bob@example.com" show_label=False%}
```
ericbroder: Suggest to not use all male gendered names for example here.

organizing_hub/templates/calls/partials/call_campaign_callers_explainer.html
```diff
@@ -0,0 +1,4 @@
...
+  Callers will be able to login and make calls after creating an account
```
ericbroder: `login` should be two words?

organizing_hub/views/call_views.py
```diff
@@ -1,17 +1,25 @@
...
 from django.utils import timezone
 from django.utils.decorators import method_decorator
-from django.views.generic import CreateView, DetailView, FormView, TemplateView
+from django.views.generic import (
+    CreateView,
+    DetailView,
+    FormView,
+    TemplateView,
+    UpdateView
+    )
```
ericbroder: fix indent?

organizing_hub/views/call_views.py
```diff
@@ -121,6 +129,46 @@ def can_make_call_for_campaign(user, call_campaign):
     return False
 
 
+def get_or_create_callers(caller_emails):
+    """
+    Get or create callers from a list of caller emails
```
ericbroder: love the new method documentation we started doing 😂

organizing_hub/views/call_views.py
```diff
@@ -121,6 +129,46 @@ def can_make_call_for_campaign(user, call_campaign):
...
+            Returns list of ints corresponding to callprofile IDs
+    """
+
+    caller_ids = []
+
+    if caller_emails is not None and caller_emails != '':
+        caller_emails = [email.strip() for email in caller_emails.split(",")]
+
+        for email in caller_emails:
+            # TODO: support multiple accounts per email
```
ericbroder: What does this mean? should there be a follow up ticket?

organizing_hub/views/call_views.py
```diff
@@ -121,6 +129,46 @@ def can_make_call_for_campaign(user, call_campaign):
...
+            (user, created) = User.objects.get_or_create(
+                email=email,
+                defaults={
+                    'username':email
```
ericbroder: Need to addemail field here, otherwise `user.email` will be blank, right?

organizing_hub/views/call_views.py
ericbroder: Suggest add more code comments to explain what is going on here. Like why do we need to create a bsd profile?

organizing_hub/views/call_views.py
```diff
@@ -121,6 +129,46 @@ def can_make_call_for_campaign(user, call_campaign):
...
+                }
+            )
+
+            if not hasattr(user,'bsdprofile'):
+                BSDProfile.objects.create(user=user)
+
+            if not hasattr(user,'callprofile'):
+                CallProfile.objects.create(user=user)
+
+            caller_ids.append(user.callprofile.id)
```
ericbroder: I get confused about this a lot, but will this return stale data sometimes? You already have the `user` object from earlier, and you just created a call profile now, so will `user.callprofile` return a call profile that was just created? might want to just actually pull the newly created call profile from the create method and return call profile id that way?

organizing_hub/views/call_views.py
```diff
@@ -292,6 +340,17 @@ def form_valid(self, form):
...
         form.instance.local_group = self.get_local_group()
         form.instance.owner = self.request.user.callprofile
+
+        """Take comma separated caller emails and get or create corresponding
+        call profiles"""
+        caller_emails = form.cleaned_data['caller_emails']
+
+        """Save call_campaign because we need a primary key in order to save
+        Many-to-many caller relationships"""
+        call_campaign = form.save()
```
ericbroder: what does `form.save()` do? im familiar with `model.save()` but not `form.save()`. why not call `super(CallCampaignCreateView, self).form_valid(form)` which is how the save was previously being called?

organizing_hub/views/call_views.py
```diff
@@ -292,6 +340,17 @@ def form_valid(self, form):
         """Set local group and owner before save"""
         form.instance.local_group = self.get_local_group()
         form.instance.owner = self.request.user.callprofile
+
+        """Take comma separated caller emails and get or create corresponding
+        call profiles"""
+        caller_emails = form.cleaned_data['caller_emails']
+
+        """Save call_campaign because we need a primary key in order to save
+        Many-to-many caller relationships"""
```
ericbroder: Are you sure about this? seems odd we need to call save twice.

organizing_hub/views/call_views.py
```diff
@@ -402,6 +461,59 @@ def post(self, request, *args, **kwargs):
...
+    slug_url_kwarg = 'uuid'
+    success_message = '''
+    Your calling campaign has been edited succesfully.
+    '''
+
+    def form_valid(self, form):
+        caller_emails = form.cleaned_data['caller_emails']
+        caller_ids = get_or_create_callers(caller_emails)
+        form.instance.callers = caller_ids
+        return super(CallCampaignUpdateView, self).form_valid(form)
```
ericbroder: Nice 👍  So this will both add and remove callers?

calls/forms.py
cjmabry: I had stopped using them because I thought you weren't for a while, let's use them from now on.

organizing_hub/templates/calls/callcampaign_detail.html
cjmabry: What do you mean exactly? This will only show if there are callers. We'll show there are 0 callers on this campaign above and there will be an edit button.

ericbroder: I'm just asking, sometimes we have conditional content for empty list.

organizing_hub/views/call_views.py
cjmabry: Fields that are used for evaluation are used outside of defaults but all fields are used when creating object. https://docs.djangoproject.com/en/2.1/ref/models/querysets/#get-or-create

organizing_hub/views/call_views.py
cjmabry: Yes.

organizing_hub/views/call_views.py
cjmabry: I don't follow. We're creating or getting a user, and then if they don't have a call profile we're creating one, and adding it to list. If they already have a call profile they'll be added to list. Are you saying if a call profile is created elsewhere while this is happening? A user should only ever have one call profile. I don't follow `might want to just actually pull the newly created call profile from the create method and return call profile id that way?`

organizing_hub/views/call_views.py
cjmabry: ```Every ModelForm also has a save() method. This method creates and saves a database object from the data bound to the form.```

https://docs.djangoproject.com/en/dev/topics/forms/modelforms/#the-save-method

We need to do this because before an object is saved we can't access the primary key because it doesn't have one. In order to save many to many relationships you have to have a primary key, or else django will throw an `instance needs to have a primary key value before a many-to-many relationship can be used.` error.

organizing_hub/views/call_views.py
cjmabry: See below.

organizing_hub/views/call_views.py
cjmabry: https://docs.djangoproject.com/en/1.11/topics/db/examples/many_to_many/

organizing_hub/views/call_views.py
cjmabry: https://ourrevolution.atlassian.net/browse/TECH-1259

Will add in code

calls/forms.py
ericbroder: At first I didn't like it, then I decided it seems useful.

organizing_hub/views/call_views.py
ericbroder: If you tested and it works, then fine. Maybe I misunderstood how this method works.

organizing_hub/views/call_views.py
ericbroder: I'm saying that when you initially get the `user` it might not have a call profile attached. So you create a call profile. But you do not clearly refresh the `user` so it might still say there is no call profile until you recheck the db?

organizing_hub/templates/calls/callcampaign_detail.html
cjmabry: Gotcha, no I think UI makes sense here since we display number of callers.

organizing_hub/views/call_views.py
ericbroder: Ok I wasn't familiar with this rule.

organizing_hub/views/call_views.py
cjmabry: So you mean do something like this?
```if not hasattr(user,'callprofile'):
                callprofile = CallProfile.objects.create(user=user)

            caller_ids.append(callprofile.id)```

Comment (ericbroder) [reply organizing_hub/views/call_views.py]: yeah basically. if it's true that user object will not auto updated without a data refresh, then need something like ```
            if hasattr(user, 'callprofile'):
                call_profile = user.callprofile
            else:
                call_profile = CallProfile.objects.create(user=user)
             caller_ids.append(callprofile.id)
```

cjmabry: In that case, I think get_or_update would work here?

ericbroder: I think you mean `get_or_create`? Yeah you could try that, but then you cant use the `user.callprofile` syntax which is convenient. But yeah essentially the code snippet above is very similar to `get_or_create`.

organizing_hub/views/call_views.py
```diff
@@ -121,6 +129,48 @@ def can_make_call_for_campaign(user, call_campaign):
...
+    """
+
+    caller_ids = []
+
+    if caller_emails is not None and caller_emails != '':
+        caller_emails = [email.strip() for email in caller_emails.split(",")]
+
+        for email in caller_emails:
+            (user, created) = User.objects.get_or_create(
+                username=email,
```
ericbroder: I just checked existing code and every other time the organizing hub looks up a user based on email, it is either `User.objects.get(email__iexact=...` or `User.objects.filter(email__iexact=...`. This means it is case insensitive, and also it is checking the email field. You are doing a case sensitive search, and checking the username field instead of the email field. Is there a reason you are doing it differently? I think probably it should just do it the same way, case insensitive, and email field, right?

cjmabry: Agreed about case insensitivity.

But email field is not unique, so we can potentially pull multiple users here if we go with just email. In django username is used as unique to filter on, and since all callers will have email address as username, I elected to query on what we know would be unique instead of what wasn't.

ericbroder: while it's possible that django allows multiple user accounts with the same email, our practice is to only have one user account per email. so i think it's fine to either just take the first result or just assume it will only have 1 match max.

not all the usernames in the db are same as email, so it's possible that searching on username for email will miss a user account that does actually have a matching email.

if you want to start supporting multiple user accounts with the same email, that is a longer conversation and i don't think we should do it now. for consistency right now it seems like we should just follow the pattern we already established which is working fine. if you want to change the pattern let's do it later after we can talk it through or research it more?

cjmabry: I'll just take first email address here. I think we use get or create for user somewhere else though, does the case insensitivity affect that one too? Should look but not near a computer atm.

ericbroder: I don't see us using get or create anywhere for User. And taking a closer look, `filter` is more of an exception, we nearly always do `User.objects.get(email__iexact=...` when looking up a hub user by email.

cjmabry: I think I used iexact initially but we wanted to use get or create to cut down on db calls, wonder if you can support case insensitivity with it. Anyways will fix.

cjmabry:
@ericbroder celebrating before QA...
http://e.lvme.me/rihw675.jpg

PR #368, opened 2018-10-19
TECH-1489 add admin page for opt out uploads

ericbroder:


contacts/models.py
```diff
@@ -184,5 +223,33 @@ def __unicode__(self):
             str(self.id)
         )
 
+    def _treat_as_none(self):
+        """Treat as None. If this record is returned from search it is likelyf a false match."""
```
cjmabry:
```suggestion
        """Treat as None. If this record is returned from search it is likely a false match."""
```

cjmabry: @ericbroder

PR #371, opened 2018-10-22
TECH-1477 Update login for callers

cjmabry:


local_groups/forms.py
```diff
@@ -95,14 +95,14 @@ class GroupLeaderSyncForm(forms.Form):
 
 
 # Customize AuthenticationForm as needed
-class GroupLoginForm(AuthenticationForm):
+class OrganizingHubLoginForm(AuthenticationForm):
```
ericbroder: Can we move this to organizing hub app too? That's really where it belongs anyway, and otherwise this name doesnt make sense inside local groups app.

local_groups/forms.py
```diff
@@ -179,7 +179,7 @@ class Meta:
 
 class PasswordResetRequestForm(AuthPasswordResetRequestForm):
```
ericbroder: This is also weird to be located in local groups app, I know it's out of scope but should be in organizing hub app.

organizing_hub/templates/partials/organizing_hub_header.html
```diff
@@ -26,8 +26,12 @@
             {% if not static_content and request.user.is_authenticated %}
             <div class="col-md-6">
               <div class="uppercase text-right">
+                {% local_group as has_group %}
```
ericbroder: The convention of `has_*` implies this is a boolean True/False value. Suggest to not use this naming convention for something that isnt a boolean.

local_groups/forms.py
cjmabry: 👍

local_groups/forms.py
cjmabry: 👍 

Can do while I'm in here.

organizing_hub/templates/partials/organizing_hub_header.html
cjmabry: :+1:

local_groups/forms.py
```diff
@@ -94,25 +86,6 @@ class GroupLeaderSyncForm(forms.Form):
     confirm = forms.BooleanField(required=True)
 
 
-# Customize AuthenticationForm as needed
-class GroupLoginForm(AuthenticationForm):
```
ericbroder: thanks!

organizing_hub/forms.py
```diff
@@ -141,6 +175,59 @@ def clean_new_password2(self):
         return password2
 
 
+class PasswordResetRequestForm(AuthPasswordResetRequestForm):
+    email = forms.EmailField(
+        label=_("Email Address"),
+        max_length=254
+    )
+
+    def get_users(self, email):
```
ericbroder: obviously out of scope, but seems like we should not be creating new user objects in a get method. maybe add a todo or follow up ticket?

organizing_hub/templates/partials/organizing_hub_header.html
```diff
@@ -26,8 +26,12 @@
             {% if not static_content and request.user.is_authenticated %}
             <div class="col-md-6">
               <div class="uppercase text-right">
+                {% local_group as group %}
+                {% if group %}
```
ericbroder: what is the purpose of the `as` statement? why not just put `{% if local_group %}`

organizing_hub/templates/partials/organizing_hub_header.html
```diff
@@ -26,8 +26,12 @@
             {% if not static_content and request.user.is_authenticated %}
             <div class="col-md-6">
               <div class="uppercase text-right">
+                {% local_group as group %}
```
ericbroder: Are we sure `local_group` is always available? This is a header used on every page. I'm concerned that this is not always in the model/context? Or might even return the wrong object in some cases? We specifically want to make sure this is the local_group associated with the logged in user. Are we confirming this somehow? For example the `group_link` does not use `local_group` form context, it uses a template tag.

organizing_hub/templates/partials/organizing_hub_header.html
cjmabry: Since local_group is a template tag you have to declare it as something first, you can't do a traditional if statement on it afaik.

organizing_hub/templates/partials/organizing_hub_header.html
cjmabry: This is a template tag that simply calls `find_local_group_by_user().` `{% group_link %}`, which was here before, essentially calls the same thing.

ericbroder: ok thanks for explaining, didn't realize this was a template tag. we should give it a better name. also would be nice to optimize and not do the same db lookup twice in this case. like either get the group once and pass it into the group link tag, or just check if the group link is none. not critical, but seems like not good performance to lookup same query twice in a row here.

PR #377, opened 2018-10-25
TECH-1389 Fix comma and spacing issue

cjmabry:


organizing_hub/views/call_views.py
```diff
@@ -147,26 +147,30 @@ def get_or_create_callers(caller_emails):
...
+        """Replace spaces and linebreaks with commas"""
+        caller_emails_stripped = caller_emails.replace(" ", ",").replace("\n", ",").replace("\r", ",")
+
+        caller_emails = [email.strip() for email in caller_emails_stripped.split(",")]
 
         for email in caller_emails:
-            """Get a user by email address if it exists. If we get multiple
-            results, grab the first."""
-            user = User.objects.filter(email__iexact=email).first()
+            if email is not '' and email is not None:
```
ericbroder: Does this work? i think `email is not ''` maybe should be `email != ''`? and why would it ever be `None`?

organizing_hub/views/call_views.py
```diff
@@ -147,26 +147,30 @@ def get_or_create_callers(caller_emails):
     caller_ids = []
 
     if caller_emails is not None and caller_emails != '':
-        caller_emails = [email.strip() for email in caller_emails.split(",")]
+        """Replace spaces and linebreaks with commas"""
+        caller_emails_stripped = caller_emails.replace(" ", ",").replace("\n", ",").replace("\r", ",")
+
+        caller_emails = [email.strip() for email in caller_emails_stripped.split(",")]
```
ericbroder: does `.strip()` do anything now? we already stripped it?

organizing_hub/views/call_views.py
cjmabry: Just being safe because of tight turnaround, I'm not sure it ever would. But this does work, tested locally, unless I'm missing something.

PR #378, opened 2018-10-25
TECH-1389 Add callers to create calling campaign page and implement edit campaign page

cjmabry:


organizing_hub/views/call_views.py
```diff
@@ -149,20 +149,16 @@ def get_or_create_callers(caller_emails):
...
 
         for email in caller_emails:
-            if email is not '' and email is not None:
+            if email is not '':
                 """Get a user by email address if it exists. If we get multiple
                 results, grab the first."""
                 user = User.objects.filter(email__iexact=email).first()
 
                 if not user:
                     user = User.objects.create(username=email, email=email)
```
ericbroder: I see another issue here. This should be `create_user` method with password set to `None`. That method will actually make sure there is an unusable pw in the db. https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#django.contrib.auth.models.UserManager.create_user

The current logic you have sets pw to `''` by default, which is not what we want.

PR #379, opened 2018-10-25
TECH-1519 service method for new user with bsd profile

ericbroder:


organizing_hub/forms.py
```diff
@@ -12,7 +12,9 @@
 from django.contrib.auth.models import User
 from django.forms import widgets
 from django.utils.translation import gettext_lazy as _
-from bsd.models import Account, BSDEvent
+from StringIO import StringIO
+from xml.etree.ElementTree import ElementTree
```
ericbroder: There was a regression when form class was moved into this app, and these imports were not moved as well.

PR #388, opened 2018-11-06
TECH-1547 Submit preselected value with modal form

cjmabry:


pages/templates/pages/tags/splash_modal.html
```diff
@@ -147,6 +147,16 @@
 
     }
 
+    // Set to 27 if the other amount input is empty
+    $('#splash-form').submit(function(event) {
+      var form = this
```
ericbroder: Please add semicolons to the end of lines that need em.

ericbroder: why do you need `form`? not used?

pages/templates/pages/tags/splash_modal.html
```diff
@@ -147,6 +147,16 @@
...
+      var amount_input = $('#splash-amount')
```
ericbroder: minor issue but i suggest namespace like `var $splashAmount = $('#splash-amount');`

PR #391, opened 2018-11-13
TECH-1466 update celery logging

ericbroder:


organizing_hub/tasks.py
```diff
@@ -397,7 +401,29 @@ def sync_contact_list_with_bsd_constituents(
 """Tasks"""
 
 
-@shared_task
+class BaseTask(Task):
```
cjmabry: This looks like same code from above, can we avoid copying/pasting code and put in central location?

cjmabry: This looks like same code from above, can we avoid copying/pasting code and put in central location?

PR #396, opened 2018-11-27
TECH-1559 update packages

ericbroder:


pages/static/dist/css/main.css
```diff
@@ -223,8 +223,8 @@ th {
   .navbar {
     display: none; }
   .btn > .caret, .form-wrapper #can_embed_form input[type=submit] > .caret,
-  .dropup > .btn > .caret, .form-wrapper #can_embed_form
```
cjmabry: Why so many changes to css?

ericbroder: @cjmabry the version upgrade for `gulp-sass` resulted in some refactors to how css is compiled. hopefully should be no regressions in UI styles.

PR #417, opened 2019-02-16
DEV-1598 let group select completed questionnaire

ericbroder:


ericbroder:
dont merge until after freeze

